PR #26, opened 2023-03-16
Shankiyani/copyright and shutdown

shan4993:
replace all `http` with `https`. Rename a helper method to Shutdown.


verbanicm (approved):

sethvargo:
I was wrong. Apparently it has to be http as codified by the robot overlords.

shan4993:
> 



> I was wrong. Apparently it has to be http as codified by the robot overlords. Sorry.

Yeah just saw that too. Can/should we change that? Otherwise just going to close this out.

PR #27, opened 2023-03-16
refactor webhookserver and pubsub cleanup func

shan4993:


verbanicm (approved):

cmd/webhook/main.go
```diff
@@ -90,7 +90,7 @@ func realMain(ctx context.Context) error {
 	shutdownCtx, done := context.WithTimeout(context.Background(), 5*time.Second)
 	defer done()
 
-	if err := webhookServer.Cleanup(); err != nil {
```
sethvargo, line 93: If you make this `Close`, then you satisfy the `io.Closer` interface, which could be helpful.

PR #30, opened 2023-03-20
add infra for retries

shan4993:


sethvargo (approved):

terraform/data/bq_views/events/unique_events.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+  delivery_id,
+  signature,
+  received event,
```
bradegler, line 4: I think you are missing a comma after received. 

Looking at the group by received appears to be a separate column. When I first read this I thought you were renaming the column e.g. `received as event` but my guess is you just missed the comma.

bradegler (changes requested):

terraform/service_retry.tf
```diff
@@ -0,0 +1,121 @@
...
+# This service is internal facing, and will only be invoked by the scheduler
+module "retry_cloud_run" {
+  source = "git::https://github.com/abcxyz/terraform-modules.git//modules/cloud_run?ref=1d5d7f3f166679b02cd3f1ec615d287d6b7002dc"
+
+  project_id = data.google_project.default.project_id
+
+  name                  = var.component_names.retry_name
+  region                = var.region
+  image                 = var.retry_image
+  ingress               = "internal"
```
verbanicm, line 92: Did you research if we can use ingress internal from cloud scheduler? Might need to be all.

terraform/service_retry.tf
```diff
@@ -0,0 +1,121 @@
...
+  secrets               = ["github-webhook-secret"]
+  service_account_email = google_service_account.retry_run_service_account.email
+  service_iam           = var.service_iam.retry
```
sqin2019, line 95: It seems the oidc service account for cloud scheduler http target need to be cloud run invoker according to [here](https://cloud.google.com/scheduler/docs/http-target-auth#set_up_the_service_account). If it will only be invoked by the scheduler, will it be better to directly grant the sa the invoker role here?

terraform/service_retry.tf
```diff
@@ -0,0 +1,121 @@
...
+      version : "latest",
+    },
+    "GITHUB_APP_ID" : {
+      name : "github-app-id",
+      version : "latest",
+    },
+    "GITHUB_WEBHOOK_ID" : {
+      name : "github-webhook-id",
+      version : "latest",
+    },
```
verbanicm, line 115: These probably dont need to be secrets.

terraform/service_retry.tf
```diff
@@ -0,0 +1,121 @@
...
+module "retry_cloud_run" {
+  source = "git::https://github.com/abcxyz/terraform-modules.git//modules/cloud_run?ref=1d5d7f3f166679b02cd3f1ec615d287d6b7002dc"
+
+  project_id = data.google_project.default.project_id
+
+  name                  = var.component_names.retry_name
+  region                = var.region
+  image                 = var.retry_image
+  ingress               = "internal"
+  secrets               = ["github-webhook-secret"]
```
verbanicm, line 93:
```suggestion
  secrets               = ["github-ssh-key", "github-app-id", "github-webhook-id"]
```

verbanicm: This array creates the secrets resource with a string "DEFAULT_VALUE", the one below addes them to the cloud run service.

terraform/service_retry.tf
shan4993: you're right it would need to be all

terraform/service_retry.tf
shan4993: true

terraform/service_retry.tf
shan4993: ah got it

terraform/service_retry.tf
shan4993: Right, this is a good idea. @verbanicm do you foresee any use case where the caller of this module would want to provide additional invokers? I wonder if explicitly calling out the Cloud Scheduler SA as the invoker here and not exposing it to callers is better. Alternatively i could do a concat..toset..and get the list of invokers that way.

verbanicm: I could see someone wanting to call the endpoint to trigger a run manually, or for debugging.

shan4993: ack will combine

terraform/variables.tf
```diff
@@ -17,37 +17,80 @@ variable "project_id" {
...
-  description = "Cloud Run service image name to deploy."
+variable "retry_image" {
+  description = "Cloud Run retry service image name to deploy."
   type        = string
   default     = "gcr.io/cloudrun/hello:latest"
 }
 
 variable "service_iam" {
-  description = "IAM member bindings for the Cloud Run service."
+  description = "IAM member bindings for the Cloud Run services."
```
verbanicm, line 70: I might split this into webhook_service_iam and retry_service_iam

terraform/variables.tf
```diff
@@ -17,37 +17,80 @@ variable "project_id" {
...
   validation {
-    condition     = can(regex("^[A-Za-z][0-9A-Za-z-]+[0-9A-Za-z]$", var.name))
+    condition     = can(regex("^[A-Za-z][0-9A-Za-z-]+[0-9A-Za-z]$", var.prefix_name))
     error_message = "Name can only contain letters, numbers, hyphens(-) and must start with letter."
   }
 }
 
-variable "domain" {
-  description = "Domain name for the Google Cloud Load Balancer."
+variable "component_names" {
```
verbanicm, line 30: I might drop this, and just hard code throughout, use the prefix_name to allow the user to provide their name, for example how the pubsub.tf components are using it.

terraform/outputs.tf
```diff
@@ -13,58 +13,71 @@
...
 output "run_service_account_email" {
   description = "Cloud Run service account email."
-  value       = google_service_account.run_service_account.email
+  value       = google_service_account.webhook_run_service_account.email
 }
 
 output "run_service_account_member" {
   description = "Cloud Run service account email iam string."
-  value       = google_service_account.run_service_account.member
+  value       = google_service_account.webhook_run_service_account.member
```
verbanicm, line 52: You could probably group these into an object and provide an additional object for the retry service, or forego the retry service but splitting it like this allows you to add the retry output later if needed without breaking the output API.

shan4993:
This PR is ready to ship but we will hold off on merging until the retry service code is written.

bradegler (approved):

PR #31, opened 2023-03-23
add config for retry

shan4993:


pkg/retry/config.go
```diff
@@ -25,12 +25,46 @@ import (
 // Config defines the set of environment variables required
 // for running the retry service.
 type Config struct {
-	Port string `env:"PORT,default=8081"`
+	AppID              string `env:"GITHUB_APP_ID,required"`
+	BigQueryID         string `env:"BIG_QUERY_ID,required"`
+	BucketURL          string `env:"BUCKET_URL,required"`
+	LockTTLClockSkewMS int    `env:"LOCK_TTL_CLOCK_SKEW_MS,required"`
+	LockTTLMinutes     int    `env:"LOCK_TTL_MINUTES,required"`
```
shan4993, line 32: i decided to change the names of these to be more descriptive, ill update the tf PR accordingly.

pkg/retry/config.go
```diff
@@ -25,12 +25,46 @@ import (
...
+	ProjectID          string `env:"PROJECT_ID,required"`
+	Port               string `env:"PORT,default=8081"`
```
yolocs, line 34: Any reason why we don't want 8080 to be the default port

pkg/retry/config.go
```diff
@@ -25,12 +25,46 @@ import (
...
+	WebhookID          string `env:"GITHUB_WEBHOOK_ID,required"`
 }
 
 // Validate validates the retry config after load.
-func (s *Config) Validate() error {
-	// TODO fill this in when Config is well defined
+func (cfg *Config) Validate() error {
+	if len(cfg.AppID) == 0 {
```
yolocs, line 40: Unless it's a slice, usually we compare with empty string.

```go
if foo == "" {
  // not valid
}
```

pkg/retry/config.go
sethvargo, line 32: We should make this `time.Duration` types instead. Envconfig will let you specify a value as "15s" and then do the conversion.

pkg/retry/config.go
sethvargo, line 40: Usually you compare against the empty string.

pkg/retry/config.go
```diff
@@ -25,12 +25,46 @@ import (
...
+		return fmt.Errorf("GITHUB_APP_ID is empty and requires a value")
```
sethvargo, line 41:
```suggestion
		return fmt.Errorf("GITHUB_APP_ID is required")
```

pkg/retry/config_test.go
```diff
@@ -0,0 +1,151 @@
...
+package retry
+
+import (
+	"testing"
+
+	"github.com/abcxyz/pkg/testutil"
+)
+
+const (
+	testAppID              = "test-app-id"
```
sethvargo, line 24: It's usually more common to hardcode these in the tests. Constants in tests would be used when you have a real external resource, like an upstream URL.

yolocs (changes requested):

pkg/retry/config.go
shan4993: The webhook service defaults to that port.

pkg/retry/config.go
shan4993: cool, im thinking about adding a default value as well. I want to make integrating with the retry service as painless as possible and that includes providing the minimal amount of configurations. On the other hand if they adjust the scheduler cadence then our default value could just be wrong. What do you think?

pkg/retry/config.go
```diff
@@ -29,6 +29,7 @@ type Config struct {
 	AppID            string        `env:"GITHUB_APP_ID,required"`
 	BigQueryID       string        `env:"BIG_QUERY_ID,required"`
 	BucketURL        string        `env:"BUCKET_URL,required"`
+	DatasetID        string        `env:"DATASET_ID,required"`
```
shan4993, line 32: Realizing ill have to pass each table name as well since they are also configurable.

pkg/retry/config.go
sethvargo: I like sane defaults.

sethvargo (approved):

bradegler (approved):

shan4993:
@yolocs good to go? Blocked on your change request.

PR #34, opened 2023-04-04
Add DLQ PubSubMessenger

shan4993:
Persist the retryLimit config to the webhook server struct.


pkg/clients/pubsub.go
```diff
@@ -24,36 +24,64 @@ import (
 
 // PubSubMessenger implements the Messenger interface for Google Cloud pubsub.
 type PubSubMessenger struct {
-	projectID string
-	topicID   string
+	projectID  string
+	topicID    string
+	topicDLQID string
```
shan4993, line 29: Should i call this `DLQTopicID`?

bradegler: I would put make this dlqTopicID and prefix the other topicID as well so its clear.

pkg/clients/pubsub.go
```diff
@@ -24,36 +24,64 @@ import (
...
 
-	client *pubsub.Client
-	topic  *pubsub.Topic
+	client   *pubsub.Client
+	topic    *pubsub.Topic
+	topicDLQ *pubsub.Topic
```
bradegler, line 33: This might be a philosophical questions, but why have a messenger that can publish to two distinctly different topics instead of just creating two separate PubSubMessenger instances with different topic IDs?

README.md
```diff
@@ -137,6 +137,7 @@ WHERE
 `PROJECT_ID`: (Required) The project where the webhook service exists in.
 `RETRY_LIMIT`: (Required) The number of retry attempts to make for failed GitHub event before writing to the DLQ.
 `TOPIC_ID`: (Required) The topic ID for PubSub.
+`TOPIC_DLQ_ID`: : (Required) The topic ID for PubSub DLQ.
```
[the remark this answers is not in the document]
shan4993: ack ill go with `DLQ_EVENTS_TOPIC_ID` and `EVENTS_TOPIC_ID`.

pkg/clients/pubsub.go
shan4993: Good point, maybe this is the preferable approach. I came across [this API doc](https://pkg.go.dev/cloud.google.com/go/pubsub#Client.Topic), specifically the part that says "Avoid creating many Topic instances if you use them to publish."

pkg/clients/pubsub.go
shan4993: bailed on this approach in favor of separate messengers for each. ty!

PR #39, opened 2023-04-26
Update webhook logic and add mock implementation for BQ

shan4993:


pkg/cli/webhook.go
```diff
@@ -87,8 +87,13 @@ func (c *WebhookServerCommand) RunUnstarted(ctx context.Context, args []string)
 	}
 	logger.Debugw("loaded configuration", "config", c.cfg)
 
-	pubsubClientOpts := append([]option.ClientOption{option.WithUserAgent("abcxyz/github-metrics-aggregator")}, c.testPubSubClientOptions...)
-	webhookServer, err := webhook.NewServer(ctx, c.cfg, pubsubClientOpts...)
+	opts := append([]option.ClientOption{option.WithUserAgent("abcxyz/github-metrics-aggregator")}, c.testPubSubClientOptions...)
```
sethvargo, line 90: As a follow-up, it would be good to set the user-agent to include a version, something like `abcxyz:github-metrics-aggregator/1.2.0`. I realize this is existing code.

pkg/webhook/webhook.go
```diff
@@ -99,6 +115,37 @@ func (s *Server) handleWebhook() http.Handler {
 
 		if err = s.eventsPubsub.Send(context.Background(), eventBytes); err != nil {
 			logger.Errorw("failed to write messages to event pubsub", "code", http.StatusInternalServerError, "body", errWritingToBackend, "error", err)
+
+			exceeds, bqQueryErr := s.bqClient.
+				FailureEventsExceedsRetryLimit(r.Context(), s.failureEventTableID, deliveryID, s.retryLimit)
```
sethvargo, line 120: Nit: It would be better to pull `r.Context()` into a `ctx` variable closer to the top.

sethvargo (approved):
(Separately) Should we consider using the renderer package for these responses?

pkg/webhook/webhook.go
shan4993: ack

shan4993:
> (Separately) Should we consider using the renderer package for these responses?

Good idea, I added https://github.com/abcxyz/github-metrics-aggregator/issues/40 so we can circle back.

pkg/cli/webhook.go
```diff
@@ -87,8 +90,19 @@ func (c *WebhookServerCommand) RunUnstarted(ctx context.Context, args []string)
 	}
 	logger.Debugw("loaded configuration", "config", c.cfg)
 
-	pubsubClientOpts := append([]option.ClientOption{option.WithUserAgent("abcxyz/github-metrics-aggregator")}, c.testPubSubClientOptions...)
-	webhookServer, err := webhook.NewServer(ctx, c.cfg, pubsubClientOpts...)
+	agent := fmt.Sprintf("abcxyz/github-metrics-aggregator/%s", version.Version)
```
shan4993, line 93: @sethvargo adding version number

pkg/webhook/database.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package webhook
+
+import "context"
+
+// DatastoreClient adheres to the interaction with a BigQuery instance.
+type DatastoreClient interface {
```
yolocs, line 20: nit: some suggestion for conciseness

You could just name it `Datastore`. You could also move this to `server.go` where the interface is used.

pkg/webhook/webhook.go
```diff
@@ -81,6 +82,22 @@ func (s *Server) handleWebhook() http.Handler {
...
+		exists, err := s.bqClient.DeliveryEventExists(ctx, s.eventsTableID, deliveryID)
+		if err != nil {
+			logger.Errorw("failed to call BigQuery", "method", "DeliveryEventExists", "code", http.StatusInternalServerError,
+				"body", errWritingToBackend, "error", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprint(w, errWritingToBackend)
+			return
+		}
+
+		// event was already processed, dont resubmit it to PubSub
```
yolocs, line 94:
```suggestion
		// event was already processed, don't resubmit it to PubSub
```

pkg/webhook/webhook.go
```diff
@@ -99,6 +116,37 @@ func (s *Server) handleWebhook() http.Handler {
 
 		if err = s.eventsPubsub.Send(context.Background(), eventBytes); err != nil {
```
yolocs, line 117:
```suggestion
		if err := s.eventsPubsub.Send(context.Background(), eventBytes); err != nil {
```

pkg/webhook/webhook.go
```diff
@@ -99,6 +116,37 @@ func (s *Server) handleWebhook() http.Handler {
...
 			logger.Errorw("failed to write messages to event pubsub", "code", http.StatusInternalServerError, "body", errWritingToBackend, "error", err)
+
+			exceeds, bqQueryErr := s.bqClient.
+				FailureEventsExceedsRetryLimit(ctx, s.failureEventTableID, deliveryID, s.retryLimit)
+			if bqQueryErr != nil {
+				logger.Errorw("failed to call BigQuery", "method", "FailureEventsExceedsRetryLimit",
+					"code", http.StatusInternalServerError, "body", errWritingToBackend, "error", bqQueryErr)
+			} else if exceeds {
+				// exceeds the limit, write to DLQ
+				if errDLQ := s.dlqEventsPubsub.Send(context.Background(), eventBytes); errDLQ != nil {
```
yolocs, line 127: Not sure why you need a different name for error
```suggestion
				if err := s.dlqEventsPubsub.Send(context.Background(), eventBytes); err != nil {
```

pkg/webhook/webhook.go
```diff
@@ -99,6 +116,37 @@ func (s *Server) handleWebhook() http.Handler {
...
+					return
+				}
+
+				// return a 200 so GitHub doesn't report a failed delivery
+				w.WriteHeader(http.StatusCreated)
+				fmt.Fprint(w, successMessage)
+				return
+			} else {
+				// record an entry in the failure events table
+				if bqWriteErr := s.bqClient.
```
yolocs, line 143: Same. `err` seems sufficient

pkg/webhook/webhook_test.go
```diff
@@ -85,51 +90,134 @@ func TestHandleWebhook(t *testing.T) {
...
+			name:                    "error_write_backend_failed_bq_failure_events_exceeds_retry_limit",
+			pubSubGRPCConn:          pubSubErrGRPCConn,
+			dlqEventsPubSubGRPCConn: dlqEventsPubSubGRPCConn,
+			payloadFile:             path.Join(testDataBasePath, "pull_request.json"),
+			payloadType:             "pull_request",
+			payloadWebhookSecret:    serverGitHubWebhookSecret,
+			expStatusCode:           http.StatusInternalServerError,
+			expRespBody:             errWritingToBackend,
+			datastoreClientOverride: &FakeDatastoreClient{failureEventsExceedsRetryLimit: &failureEventsExceedsRetryLimitRes{res: false, err: errors.New("error")}},
+			// 152
```
yolocs, line 184: What are these numbers?

shan4993:
This PR is held up on a missing BQ permission for `bigquery.jobs.create` which includes queries. This is causing the integ test to fail.

pkg/webhook/webhook_test.go
shan4993: slipped in, ill remove. They were to confirm line coverage in the unit test.

pkg/cli/webhook.go
sethvargo, line 93:
```suggestion
	agent := fmt.Sprintf("abcxyz:github-metrics-aggregator/%s", version.Version)
```

The format for user-agent strings is structured and there can only be a single slash separating the name and version.

sethvargo (approved):

pkg/webhook/mock.go
```diff
@@ -0,0 +1,62 @@
...
+	res bool
+	err error
+}
+
+type failureEventsExceedsRetryLimitRes struct {
+	res bool
+	err error
+}
+
+type FakeDatastore struct {
```
yolocs, line 29: Could be a follow-up - This says fake but really is a mock. The difference is:

- Fake: A realistic but naive implementation (e.g. in memory storage). Should be unit tested. Often exposed.
- Mock: Not a real implementation. Don't require unit test. Only used in local unit test.

yolocs (approved):

pkg/webhook/mock.go
shan4993: Thanks for sharing going to update now.

PR #44, opened 2023-04-29
allow multiple revisions to be created and tested in ci at the same time

verbanicm:


.github/workflows/ci.yml
```diff
@@ -33,12 +33,12 @@ env:
   INTEGRATION_WEBHOOK_PROJECT_ID: 'github-metrics-aggreg-i-64d426'
   INTEGRATION_WEBHOOK_REGION: 'us-central1'
   INTEGRATION_WEBHOOK_SERVICE_NAME: 'github-metrics-webhook-2578'
-  INTEGRATION_WEBHOOK_SERVICE_AUDIENCE: 'https://github-metrics-webhook-2578-u4iwdc42oa-uc.a.run.app'
+  INTEGRATION_WEBHOOK_URL: 'https://${{ github.sha }}--github-metrics-webhook-2578-u4iwdc42oa-uc.a.run.app'
```
shan4993, line 36: Double dash intentional?

shan4993 (approved):

.github/workflows/ci.yml
verbanicm: Yea, thats how cloud run references tagged revisions.

terraform/variables.tf
```diff
@@ -27,6 +27,12 @@ variable "prefix_name" {
   }
 }
 
+variable "enable_webhook_gclb" {
```
sethvargo, line 30: Why wouldn't we always just use the gclb? I'm generally opposed to conditionals that change the flow of Terraform because it leads to spaghetti code that becomes impossible to reason about.

sethvargo (changes requested):

terraform/variables.tf
verbanicm: I cant target a tagged cloud run revision URL from the GCLB, so for CI, I am disabling it.

I didnt want to create a whole separate module just for CI.

verbanicm: To elaborate, I am trying to allow multiple PRs to create revisions for a cloud run services, this requires tagging the revision and hitting it by that specific URL.

verbanicm: We could just create it and not use it and let the config pass cloud run ingress and let it be changed to all and just the run url directly. Im open to ideas

sethvargo: Interesting - we can't route to a specific revision based on a header or query param?

Let's just document it very very clearly inline.

verbanicm: Let me research that first, if not ill document.

verbanicm: So this COULD be done with a serverless neg URL mask. We arent setup for this and would take a bit of refactoring. I will create an issue and document the current design inline.

yolocs: > I didnt want to create a whole separate module just for CI.

If cloud run is all you need, this is definitely worth trying. But to call out - because revision is only a cloud run thing, you still can't avoid the problem if you need other isolated resources for parallel CI runs. What we've been doing in JVS/pmap is to decompose the e2e setup module into 1) more static resources 2) resources needed isolation. That way, in CI we can still reuse the 2nd module.

verbanicm:
@sethvargo I'd like to get this merged to unblock current development. We can improve this later. I created https://github.com/abcxyz/github-metrics-aggregator/issues/45 to track.

terraform/variables.tf
verbanicm: Sounds good, I created https://github.com/abcxyz/github-metrics-aggregator/issues/45 to track this

sethvargo (approved):

PR #50, opened 2023-05-12
add logs and allow configurabe log level and mode

shan4993:


pkg/retry/retry.go
```diff
@@ -61,7 +62,12 @@ func (s *Server) handleRetry() http.Handler {
 			return
 		}
 
-		for cursor != "" {
+		logger.Infow("retrieved last checkpoint", "checkpoint", cursor)
+
+		// for each run track the total number of events retrieve and failure events
+		totalEventCount, failedEventCount := 0, 0
+
+		for ok := true; ok; ok = (cursor != "") {
```
shan4993, line 70: @sethvargo i brought this back because we need the loop to run at least once. The first time this service is invoked, the checkpoint will be empty. I'm attempting to replicate a do while loop in go. If you agree, is this the correct way?

verbanicm, line 70: This doesnt seem right? Why not just cursor != ""

verbanicm: Ill defer to your comment for Seth for the idiomatic approach.

sethvargo: Yea this is common. The other option is to `break` under a condition, but this is fine.

pkg/retry/retry.go
```diff
@@ -61,7 +62,12 @@ func (s *Server) handleRetry() http.Handler {
 			return
 		}
 
-		for cursor != "" {
+		logger.Infow("retrieved last checkpoint", "checkpoint", cursor)
+
+		// for each run track the total number of events retrieve and failure events
+		totalEventCount, failedEventCount := 0, 0
```
sethvargo, line 68: Nit, it's more idiomatic to do:

```go
var totalEventCount int
var failedEventCount int
```

pkg/retry/retry.go
```diff
@@ -103,7 +119,8 @@ func (s *Server) handleRetry() http.Handler {
 			}
 		}
 
-		logger.Infow("successful", "code", http.StatusAccepted, "body", acceptedMessage)
+		logger.Infow("successful", "code", http.StatusAccepted, "body", acceptedMessage, "totalEventCount", totalEventCount,
```
sethvargo, line 122: Nit: when the log gets long, split it to KV pairs:

```suggestion
		logger.Infow("successful",
			"code", http.StatusAccepted,
			"body", acceptedMessage,
			"totalEventCount", totalEventCount,
```

pkg/retry/retry.go
```diff
@@ -83,18 +94,23 @@ func (s *Server) handleRetry() http.Handler {
...
+
+				// append to the total failure events counter
+				failedEventCount += 1
 			}
 
-			// periodic update to checkpoint
-			if err := s.datastore.WriteCheckpointID(ctx, s.checkpointTableID, "checkpointID", time.Now().String()); err != nil {
+			logger.Infow("write latest checkpoint", "checkpoint", cursor)
+
+			if err := s.datastore.WriteCheckpointID(ctx, s.checkpointTableID, cursor, time.Now().String()); err != nil {
```
sethvargo, line 113: 1. Please grab a copy of `time.Now()` at the very start of the invocation, otherwise you'll have races.

2. Please use `UTC` of every and all times.

3. Is `String` the correct function? I would expect `.Format()` in a specific format instead.

terraform/variables.tf
```diff
@@ -231,7 +231,7 @@ variable "lock_ttl_clock_skew" {
 }
 
 variable "cloud_scheduler_deadline_duration" {
-  description = "The deadline for job attempts in seconds. If the request handler does not respond by this deadline then the request is cancelled and the attempt is marked as a DEADLINE_EXCEEDED failure. Defaults to 30 minutes (max)"
+  description = "The deadline for job attempts in seconds. If the request handler does not respond by this deadline then the request is cancelled and the attempt is marked as a DEADLINE_EXCEEDED failure. Default is 30 minutes (max)"
```
sethvargo, line 234: Why this change? I think this is no longer grammatically correct...

terraform/variables.tf
```diff
@@ -274,3 +274,15 @@ variable "region" {
   type        = string
   default     = "us-central1"
 }
+
+variable "log_mode" {
+  description = "The mode for logging. Default log mode is production."
```
sethvargo, line 279:
```suggestion
  description = "The mode for logging. Defaults to 'production'."
```

terraform/variables.tf
```diff
@@ -274,3 +274,15 @@ variable "region" {
...
+  type        = string
+  default     = "production"
+}
+
+variable "log_level" {
+  description = "The level for logging. Default log level is warning."
```
sethvargo, line 285:
```suggestion
  description = "The level for logging. Defaults to 'warning'."
```

terraform/variables.tf
verbanicm: It previously has a typo `Default is to` so I suggested `Default is`, but I think you're right.

`Defaults to 30 minutes`

verbanicm (approved):

pkg/retry/retry.go
```diff
@@ -37,41 +37,65 @@ const (
 // events.
 func (s *Server) handleRetry() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		now := time.Now().UTC().Format(time.UnixDate)
```
sethvargo, line 40: Sorry, nit: I'd make this `time.Now().UTC()` so you have a real time object that you can use throughout.

Then when you need to put it in the query, call `.Format(time.UnixDate)`.

sethvargo (approved):

PR #52, opened 2023-05-12
throw 500 when GCS has an unexpected error

shan4993:


pkg/retry/retry.go
shan4993, line 1: I suspect there is an unknown failure with GCS which is causing a silent exit. If this is correct, this change should reveal the error.

sethvargo (approved):

sethvargo:
In the future, we'll want to also emit a metric so we can alert on these kinds of things, but that's a day 2 problem :)

PR #53, opened 2023-05-12
rename BUCKET_URL to BUCKET_NAME

shan4993:


terraform/service_retry.tf
```diff
@@ -100,7 +100,7 @@ module "retry_cloud_run" {
   }
   envvars = {
     "BIG_QUERY_PROJECT_ID" : var.bigquery_project_id,
-    "BUCKET_URL" : google_storage_bucket.retry_lock.url,
+    "BUCKET_NAME" : google_storage_bucket.retry_lock.name,
```
shan4993, line 103: i believe the gcslock package wants the name of the bucket and not its URL.

pkg/retry/config.go
```diff
@@ -128,9 +128,9 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 
 	f.StringVar(&cli.StringVar{
 		Name:   "bucket-url",
-		Target: &cfg.BucketURL,
-		EnvVar: "BUCKET_URL",
-		Usage:  `The URL for the bucket that holds the lock to enforce synchronous processing of the retry service.`,
+		Target: &cfg.BucketName,
+		EnvVar: "BUCKET_NAME",
+		Usage:  `The name of the bucket that holds the lock to enforce synchronous processing of the retry service.`,
```
sethvargo, line 133: Maybe add the `Example` key?

sethvargo (approved):

pkg/retry/config.go
shan4993: ack

drevell (approved):

PR #63, opened 2023-05-16
refactor retry logic

shan4993:


pkg/retry/retry.go
```diff
@@ -82,17 +83,24 @@ func (s *Server) handleRetry() http.Handler {
...
 		// ensure we run the loop at least once
 		for ok := true; ok; ok = (cursor != "") {
-			// call list deliveries API
-			deliveries, res, err := s.github.ListDeliveries(ctx, &github.ListCursorOptions{Cursor: cursor})
+			// call list deliveries API, first call is intentionally an empty string
+			deliveries, res, err := s.github.ListDeliveries(ctx,
+				&github.ListCursorOptions{
+					Cursor:  cursor,
+					PerPage: 100,
+				})
```
sethvargo, line 103: Style nit:

```suggestion
			deliveries, res, err := s.github.ListDeliveries(ctx, &github.ListCursorOptions{
				Cursor:  cursor,
				PerPage: 100,
			})
```

pkg/retry/retry.go
```diff
@@ -104,61 +112,80 @@ func (s *Server) handleRetry() http.Handler {
...
-				logger.Errorw("failed to call WriteCheckpointID",
-					"code", http.StatusInternalServerError,
-					"body", errWriteCheckpoint,
-					"method", "RedeliverEvent",
-					"error", err)
-				w.WriteHeader(http.StatusInternalServerError)
-				fmt.Fprint(w, http.StatusText(http.StatusInternalServerError))
-				return
+				if err := s.datastore.WriteCheckpointID(ctx, s.checkpointTableID,
+					strconv.FormatInt(*newCheckpoint, 10), now.Format(time.DateTime)); err != nil {
```
sethvargo, line 167: Since we're refactoring this: maybe name these variables before inserting them into the function call?

pkg/retry/retry.go
```diff
@@ -104,61 +112,80 @@ func (s *Server) handleRetry() http.Handler {
...
+					logger.Errorw("failed to call WriteCheckpointID",
+						"code", http.StatusInternalServerError,
+						"body", errWriteCheckpoint,
+						"method", "RedeliverEvent",
+						"error", err,
+						"totalEventCount", totalEventCount,
+						"failedEventCount", failedEventCount,
+					)
+					w.WriteHeader(http.StatusInternalServerError)
+					fmt.Fprint(w, http.StatusText(http.StatusInternalServerError))
```
sethvargo, line 177: Prefer `http.Error` instead.

sethvargo (approved):

pkg/retry/retry.go
```diff
@@ -104,61 +112,80 @@ func (s *Server) handleRetry() http.Handler {
...
 				}
 
+				// append to the total failure events counter
+				failedEventCount += 1
+
 				logger.Infow("redeliver failed event", "eventID", *event.ID, "guid", *event.GUID)
 
 				if err := s.github.RedeliverEvent(ctx, *event.ID); err != nil {
-					logger.Errorw("failed to call RedeliverEvent",
+					logger.Errorw("failed to call RedeliverEvent, stop processing",
```
verbanicm, line 145: Should we consider rate limiting here? Just curious about downtime and we have to send a ton of these, we could hit rate limits and need to throttle ourselves. Rather than failing completely we could back off (or check headers for the next time we can send).

This could be a follow up.

verbanicm (approved):

pkg/retry/retry.go
shan4993: nice, updated

pkg/retry/retry.go
shan4993: Good call, it looks like we can anticipate a [422 from the redeliver API](https://docs.github.com/en/rest/orgs/webhooks?apiVersion=2022-11-28#redeliver-a-delivery-for-an-organization-webhook). Created an issue for this https://github.com/abcxyz/github-metrics-aggregator/issues/64.

pkg/retry/retry_test.go
```diff
@@ -55,6 +56,7 @@ func TestHandleRetry(t *testing.T) {
 				listDeliveries: &listDeliveriesRes{
 					deliveries: []*github.HookDelivery{
 						{
+							ID:         toPtr[int64](101),
```
sethvargo, line 59: I actually don't know which is more idiomatic since generics are relatively new, but you should also be able to:

```go
toPtr(int64(101))
```

PR #66, opened 2023-05-17
record the new checkpoint

shankiyani:


pkg/retry/retry.go
```diff
@@ -186,22 +189,19 @@ func (s *Server) handleRetry() http.Handler {
 
 			// every event from newCheckpoint to lastCheckpoint has been processed,
 			// overwrite the checkpoint
-			if processed {
```
shan4993, line 189: this variable is not necessary, if we get this far we should update the checkpoint

pkg/retry/retry.go
```diff
@@ -88,9 +88,8 @@ func (s *Server) handleRetry() http.Handler {
 		// for each run track the total number of events retrieve and failure events
 		var totalEventCount int
 		var failedEventCount int
-		var newCheckpoint *int64
+		var newCheckpoint string
```
sethvargo, line 91: Whats the reason for the type change?

pkg/retry/retry.go
```diff
@@ -113,6 +112,11 @@ func (s *Server) handleRetry() http.Handler {
 
 			logger.Infow("retrieve deliveries from GitHub", "cursor", cursor, "size", len(deliveries))
 
+			// store the ID of the first delivery from the first list call
+			if newCheckpoint == "" {
+				newCheckpoint = strconv.FormatInt(*deliveries[0].ID, 10)
```
verbanicm, line 117: Will the next page include this checkpoint or get everything AFTER it?

pkg/retry/retry.go
shan4993: we store checkpoints as string, it felt more consistent.

pkg/retry/retry.go
shan4993: the former, it will begin with the provided cursor.

pkg/retry/retry.go
yolocs: It seems in BQ the checkpoint is a string and `lastCheckpoint` is also a string. So sooner or later we need to convert the delivery id to a string.

yolocs (approved):

pkg/retry/retry.go
shan4993: our first list call will always start without a cursor since the events will be in descending time order.

pkg/retry/retry.go
shan4993: right, it might be confusing referring to checkpoints (newCheckpoint and lastCheckpoint) as two different types.

PR #67, opened 2023-05-17
limit writes to checkpoint

shan4993:


bradegler (approved):

pkg/retry/retry.go
shan4993, line 1: The idea here is to update the checkpoint less often. Given that the [GitHub API ](https://docs.github.com/en/rest/apps/webhooks?apiVersion=2022-11-28#list-deliveries-for-an-app-webhook) does not allow us to return events in ascending order, we must start from the most recent deliveries (passing an empty cursor) and start paginating until we get to the `lastCheckpoint`. If we update the checkpoint to the `newCheckpoint` we are indicating that all events older than the `newCheckpoint` have been ACK'd. Instead this change will only update the checkpoint if 

a) the `lastCheckpoint` was reached
b) we exhausted processing of all events aka we reached the end.

PR #68, opened 2023-05-17
load failed events into memory

shan4993:


pkg/retry/retry.go
```diff
@@ -35,6 +35,11 @@ const (
 	errCallingGitHub       = "Failed to call GitHub."
 )
 
+type EventIdentifier struct {
```
verbanicm, line 38: doc

pkg/retry/retry.go
```diff
@@ -141,52 +143,58 @@ func (s *Server) handleRetry() http.Handler {
...
-						logger.Infow("skipping redeliver event because it has already been processed", "eventID", *event.ID, "guid", *event.GUID)
-						continue
-					}
-
-					logger.Errorw("failed to call RedeliverEvent, stop processing",
+				failedEventsHistory = append(failedEventsHistory, &EventIdentifier{eventID: event.ID, guid: *event.GUID})
+			}
+		}
+
+		for i := len(failedEventsHistory) - 1; len(failedEventsHistory) > 0 && i >= 0; i-- {
```
verbanicm, line 150: doc this whole thing so we know y we are using the failure list and going backwards

verbanicm (approved):

pkg/retry/retry.go
```diff
@@ -35,6 +35,13 @@ const (
 	errCallingGitHub       = "Failed to call GitHub."
 )
 
+// EventIdentifier represents the required information used by the retry
+// service for handling a GitHub event.
+type EventIdentifier struct {
+	eventID *int64
```
sethvargo, line 41: Why a pointer?

pkg/retry/retry.go
```diff
@@ -35,6 +35,13 @@ const (
 	errCallingGitHub       = "Failed to call GitHub."
 )
 
+// EventIdentifier represents the required information used by the retry
+// service for handling a GitHub event.
+type EventIdentifier struct {
```
yolocs, line 40: nit: Really need to be exposed?

pkg/retry/retry.go
```diff
@@ -141,52 +146,61 @@ func (s *Server) handleRetry() http.Handler {
...
-						return
-					}
-
-					if exists {
-						logger.Infow("skipping redeliver event because it has already been processed", "eventID", *event.ID, "guid", *event.GUID)
-						continue
-					}
-
-					logger.Errorw("failed to call RedeliverEvent, stop processing",
+				failedEventsHistory = append(failedEventsHistory, &EventIdentifier{eventID: event.ID, guid: *event.GUID})
```
yolocs, line 149: Is `event` a pointer?
```suggestion
				failedEventsHistory = append(failedEventsHistory, &EventIdentifier{eventID: event.ID, guid: event.GUID})
```

pkg/retry/retry.go
```diff
@@ -141,52 +146,61 @@ func (s *Server) handleRetry() http.Handler {
...
+		// work backwards from the list of failed events then attempt redelivery and
+		// increment the newCheckpoint in an effort to close the gap to the most
+		// recent event, this should alleviate pressure on future runs
+		for i := len(failedEventsHistory) - 1; len(failedEventsHistory) > 0 && i >= 0; i-- {
+			eventIdentifier := failedEventsHistory[i]
+			logger.Infow("redeliver failed event", "eventID", eventIdentifier.eventID)
+
+			if err := s.github.RedeliverEvent(ctx, *eventIdentifier.eventID); err != nil {
+				var acceptedErr *github.AcceptedError
+				if errors.As(err, &acceptedErr) {
```
sethvargo, line 162: Do we need to remove the entry from `failedEventsHistory` in this case?

pkg/retry/retry.go
```diff
@@ -141,52 +146,61 @@ func (s *Server) handleRetry() http.Handler {
...
 						"error", err,
-						"totalEventCount", totalEventCount,
-						"failedEventCount", failedEventCount,
 					)
 					http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 					return
 				}
+
+				if exists {
+					logger.Infow("skipping redeliver event because it has already been processed", "eventID", *eventIdentifier.eventID, "guid", eventIdentifier.guid)
```
yolocs, line 185:
```suggestion
					logger.Infow("skipping redeliver event because it has already been processed", "eventID", eventIdentifier.eventID, "guid", eventIdentifier.guid)
```

pkg/retry/retry.go
```diff
@@ -141,52 +146,61 @@ func (s *Server) handleRetry() http.Handler {
...
+		// recent event, this should alleviate pressure on future runs
+		for i := len(failedEventsHistory) - 1; len(failedEventsHistory) > 0 && i >= 0; i-- {
+			eventIdentifier := failedEventsHistory[i]
+			logger.Infow("redeliver failed event", "eventID", eventIdentifier.eventID)
+
+			if err := s.github.RedeliverEvent(ctx, *eventIdentifier.eventID); err != nil {
+				var acceptedErr *github.AcceptedError
+				if errors.As(err, &acceptedErr) {
+					logger.Infow("skipping redeliver event because it has already been submitted to GitHub",
+						"eventID", *eventIdentifier.eventID,
```
yolocs, line 164: Throughout, I'd double check whether you're use `*` unnecessarily
```suggestion
						"eventID", eventIdentifier.eventID,
```

pkg/retry/retry.go
```diff
@@ -141,52 +146,61 @@ func (s *Server) handleRetry() http.Handler {
...
-						"guid", *event.GUID,
+						"body", errDeliveryEventExists,
 						"error", err,
-						"totalEventCount", totalEventCount,
-						"failedEventCount", failedEventCount,
 					)
 					http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 					return
 				}
+
```
sethvargo, line 183: Nit: cuddle this since it's related to the logic on L171

```suggestion
```

pkg/retry/retry.go
```diff
@@ -141,52 +146,61 @@ func (s *Server) handleRetry() http.Handler {
...
+					"code", http.StatusInternalServerError,
+					"body", errCallingGitHub,
+					"method", "RedeliverEvent",
+					"guid", eventIdentifier.guid,
+					"error", err,
+					"totalEventCount", totalEventCount,
+					"failedEventCount", failedEventCount,
+				)
+				http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+				return
```
sethvargo, line 199: Why stop processing on the first failed re-delivery? I think we might want more of a heuristic (like stop after 5 failures).

yolocs (commented):
Just so you know this function starts to get unreasonably long... Should consider a refactor as a follow up

pkg/retry/retry.go
shan4993: we are cautiously moving the checkpoint, as soon as we experience a failure we want to stop so the next run will start from the last successful processed event.

pkg/retry/retry.go
shan4993: the github library im using is using a *int64 so i was just matching them.

pkg/retry/retry.go
yolocs: OK, I didn't pay attention to the value we got from GH. Consider de-pointer them in the EventIdentifier to make the code here (and other places) cleaner

pkg/retry/retry.go
shan4993: it is a pointer of this following type `*github.HookDelivery`

pkg/retry/retry.go
shan4993: it was a pointer but i removed it.

pkg/retry/retry.go
shan4993: ill remove it though

PR #69, opened 2023-05-17
dont skip the update to the new checkpoint

shan4993:


pkg/retry/retry.go
shan4993, line 1: the diff is unclear but im basically removing the `continue`  and converting to an if/else. There is a bug where we hit the continue and never update the `newCheckpoint`.

pkg/retry/retry.go
```diff
@@ -165,37 +165,36 @@ func (s *Server) handleRetry() http.Handler {
...
 					)
-					continue
-				}
-
-				// found an unaccepted error, check if its already in the events table
-				exists, err := s.datastore.DeliveryEventExists(ctx, s.eventsTableID, eventIdentifier.guid)
-				if err != nil {
-					logger.Errorw("failed to call BigQuery",
-						"method", "DeliveryEventExists",
+				} else {
```
verbanicm, line 168: Should this just be `if !errors.As(err, &acceptedErr) {` insteaf of an if/else? Might clean up the look of it. IE only check if in table if we get an unknown error

verbanicm (approved):

pkg/retry/retry.go
shan4993: ack agreed

verbanicm (approved):

PR #70, opened 2023-05-17
write next checkpoint as the final step

shan4993:
retrieving the latest checkpointID based on createdAt time is not sufficient given our burst like writes to BQ. This PR aims to track the new checkpoint and write it at the very end before returning.

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -15,6 +15,7 @@
 package retry
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


pkg/retry/retry.go
```diff
@@ -221,3 +219,25 @@ func (s *Server) handleRetry() http.Handler {
 		fmt.Fprint(w, http.StatusText(http.StatusAccepted))
 	})
 }
+
+// writeMostRecentCheckpoint is a helper function to write to the checkpoint
+// table with the last successfully processed checkpoint denoted by
+// newCheckpoint.
+func (s *Server) writeMostRecentCheckpoint(ctx context.Context, w http.ResponseWriter,
```
sethvargo, line 226: Nit: pull the logger from the context instead.

shan4993: ack done

verbanicm (approved):

PR #72, opened 2023-05-18
stop processing additional pages when checkpoint is found

shan4993:
+ track the redeliveredEventcount
+ less logging noise

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -93,18 +93,19 @@ func (s *Server) handleRetry() http.Handler {
 
 		logger.Infow("retrieved last checkpoint", "lastCheckpoint", lastCheckpoint)
 
-		// for each run track the total number of events retrieve and failure events
 		var totalEventCount int
 		var failedEventCount int
+		var redeliveredEventCount int
 		var newCheckpoint string
 		var cursor string
 
 		// store all observed failures in memory from the latest event up to the lastCheckpoint
 		var failedEventsHistory []*eventIdentifier
+		var found bool
 
 		// the first run of this service will not have a cursor therefore we must
 		// ensure we run the loop at least once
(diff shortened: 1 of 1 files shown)


pkg/retry/retry.go
```diff
@@ -93,18 +93,19 @@ func (s *Server) handleRetry() http.Handler {
 
 		logger.Infow("retrieved last checkpoint", "lastCheckpoint", lastCheckpoint)
 
-		// for each run track the total number of events retrieve and failure events
 		var totalEventCount int
 		var failedEventCount int
+		var redeliveredEventCount int
```
shan4993, line 98: this will always be `<= failedEventCount`

pkg/retry/retry.go
```diff
@@ -93,18 +93,19 @@ func (s *Server) handleRetry() http.Handler {
...
 		var cursor string
 
 		// store all observed failures in memory from the latest event up to the lastCheckpoint
 		var failedEventsHistory []*eventIdentifier
+		var found bool
 
 		// the first run of this service will not have a cursor therefore we must
 		// ensure we run the loop at least once
-		for ok := true; ok; ok = (cursor != "") {
+		for ok := true; ok; ok = (cursor != "" && !found) {
```
shan4993, line 108: currently we are exhaustively calling for all deliveries, this should stop that unintended behavior

pkg/retry/retry.go
```diff
@@ -201,18 +202,21 @@ func (s *Server) handleRetry() http.Handler {
 				}
 			}
 
+			logger.Infow("detected a failed event and successfully redelivered", "eventID", eventIdentifier.eventID)
```
shan4993, line 205: too much logging noise, moving this down so we log only when we successfully redeliver a failure after the fact.

verbanicm (approved):

PR #73, opened 2023-05-18
initial new checkpoint should be the lastCheckpoint

shan4993:


sethvargo (approved):

pkg/retry/retry.go
```diff
@@ -96,8 +96,8 @@ func (s *Server) handleRetry() http.Handler {
 		var totalEventCount int
 		var failedEventCount int
 		var redeliveredEventCount int
-		var newCheckpoint string
 		var cursor string
+		newCheckpoint := lastCheckpoint
```
shan4993, line 100: bug observed where the checkpoint is set to an empty string, causing reprocessing of the entire event history.

PR #75, opened 2023-05-22
feat: updates to leech pipeline for deployment

bradegler:
* Converted to use the CLI from pkg to get clean flag parsing. Dataflow has no support for env variables.
* Changed to allow GitHub attributes to be read from secret manager. Dataflow has no support for secret injection.
* Added `job_name` column to leech status table to provide link back to Dataflow job. JOB_NAME is an env variable that is provided by Dataflow.


pkg/cli/leech.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (c *LeechCommand) Desc() string {
+	return `Start a leech pipeline for GitHub Metrics Aggregator`
+}
+
+func (c *LeechCommand) Help() string {
+	return `
+Usage: {{ COMMAND }} [options]
+  Start a leech pipeline for GitHub Metrics Aggregator.
```
sethvargo, line 47:
```suggestion

  Start a leech pipeline for GitHub Metrics Aggregator.
```

terraform/main.tf
```diff
@@ -23,6 +23,7 @@ resource "google_project_service" "default" {
     "pubsub.googleapis.com",
     "storage.googleapis.com",
     "cloudscheduler.googleapis.com",
+    "dataflow.googleapis.com",
```
sethvargo, line 26: Nit: sort these

sethvargo (approved):

PR #82, opened 2023-05-27
Add logging metric for event count

shan4993:


terraform/service_retry.tf
```diff
@@ -137,3 +137,44 @@ module "retry_cloud_run" {
...
+resource "google_logging_metric" "event_count_metric" {
+  project = data.google_project.default.project_id
+
+  name        = "Retry Service Events"
+  description = "Count of total, failure, and redelivered events per retry service invocation"
+  filter      = <<-EOT
+    resource.type="cloud_run_revision" AND
+    resource.labels.service_name="${module.retry_cloud_run.service_name}" AND
+    resource.labels.location="${module.retry_cloud_run.region}" AND 
+    jsonPayload.message=successful AND
```
shan4993, line 151: source log come from https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/retry/retry.go#L223-L229

sethvargo:
@shan4993 we should talk about this a bit more. The code is fine, but there's a chicken-and-egg problem with having the IaC tooling register metrics (vs the app). For example, you may notice that EN has a dedicated service for registering metrics which is tightly coupled to the code. This was because otherwise you can accidentally create a circular dependency between the IaC and the app.

verbanicm (dismissed):

shan4993:
> @shan4993 we should talk about this a bit more. The code is fine, but there's a chicken-and-egg problem with having the IaC tooling register metrics (vs the app). For example, you may notice that EN has a dedicated service for registering metrics which is tightly coupled to the code. This was because otherwise you can accidentally create a circular dependency between the IaC and the app.

This is fair, i wasnt sure what would happen if the metric is created but the logs don't exist yet. I assumed it would just be an empty graph but maybe that assumption is incorrect.

PR #85, opened 2023-06-01
fix: updated to use correct wif configuration

bradegler:


.github/workflows/release.yml
```diff
@@ -45,8 +43,8 @@ jobs:
         name: 'Authenticate to Google Cloud'
         uses: 'google-github-actions/auth@c4799db9111fba4461e9f9da8732e5057b394f72' # ratchet:google-github-actions/auth@v0
         with:
-          workload_identity_provider: '${{ env.WIF_PROVIDER }}'
-          service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
+          workload_identity_provider: '${{ vars.WIF_PROVIDER }}'
```
verbanicm, line 46: Should we make this the production environment also?

bradegler: oh, I missed those, I'll update them. I'm eventually going to rip prod deployment out of this workflow and into infra.

bradegler: oh, I missed those, I'll update them. I'm eventually going to rip prod deployment out of this workflow and into infra.

verbanicm: can you do it now so we dont break existing production?

verbanicm: also, theres probably some IAM permissions in infra needed here.

bradegler: great, more PRs! :)

verbanicm (approved):

PR #98, opened 2023-06-22
Add views for Pull Request data

lock14:
* Refactored `unique_events.sql` to be located in the `base` directory as it is the base view upon which all other views will be derived.
* Added `pull_request_events.sql` which creates a view that filters events to those only related to Pull Requests.
* Added `pull_requests.sql` which provides a view of distinct pull requests.


terraform/bigquery.tf
```diff
@@ -312,7 +312,31 @@ resource "google_bigquery_table_iam_member" "failure_events_viewers" {
   member     = each.value
 }
 
-// Unique Events View 
+// Base Views
+
+resource "google_bigquery_table" "base_views" {
+  for_each = fileset("${path.module}/data/bq_views/base", "*")
```
sethvargo, line 318: Should this be `*.sql`?

terraform/data/bq_views/events/pull_request_events.sql
```diff
@@ -0,0 +1,64 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- Filters events to those just pertaining to pull requests and parses relevant
+-- pull request fields into columns.
+--
+-- Relevaant Github Documentation:
```
sethvargo, line 18:
```suggestion
-- Relevaant GitHub Documentation:
```

terraform/data/bq_views/events/pull_request_events.sql
```diff
@@ -0,0 +1,64 @@
...
+  JSON_VALUE(PARSE_JSON(payload), "$.pull_request.state") state,
+  JSON_VALUE(PARSE_JSON(payload), "$.pull_request.title") title,
+  TIMESTAMP(JSON_VALUE(PARSE_JSON(payload), "$.pull_request.updated_at")) updated_at,
+  INT64(JSON_QUERY(PARSE_JSON(payload), "$.pull_request.user.id")) author_id,
+  JSON_VALUE(PARSE_JSON(payload), "$.pull_request.user.login") author,
+  TIMESTAMP_DIFF( TIMESTAMP(JSON_VALUE(PARSE_JSON(payload), "$.pull_request.closed_at")), TIMESTAMP(JSON_VALUE(PARSE_JSON(payload), "$.pull_request.created_at")), SECOND) open_duration_seconds
+FROM
+  `${dataset_id}.${table_id}`
+WHERE
+  event = "pull_request";
```
sethvargo, line 64: I'll defer to @bradegler, but I think this is quite inefficient because we end up parsing the JSON 50 times. It might be better to parse the JSON as a CTE and then do a sub-SELECT.

verbanicm: You dont need the parse, you can just do `JSON_VALUE(payload, "$.action") action`

lock14: I was hoping query engine was smart enough to only do `PARSE_JSON(payload)` once even though it is stated several times. I think mike is write that it is not strictly needed though.

sethvargo: I just want to make sure `JSON_VALUE` isn't going to parse the substantially large payload N times for each row.

lock14: `JSON_VALUE` is not sufficient as it returns a string https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_value. 

We don't want every column to be a string, but parsing funcitions such as `INT64` take a a `json_expr` and not a string. So we do actutally need to parse the JSON.

terraform/data/bq_views/events/pull_request_events.sql
lock14: fixed.

terraform/bigquery.tf
lock14: yes, fixed.

terraform/data/bq_views/events/pull_request_events.sql
lock14: I updated to parse the json in a subquery.

verbanicm: I dont think we need to do this, JSON_VALUE is efficient at getting the values out. If you need to cast to another type you can do the following for the type you want, specifically INT64 and BOOL:

```
  SAFE_CAST(JSON_VALUE(payload, "$.organization.id") AS INT64) organization_id
```

terraform/bigquery.tf
verbanicm: I dont think you committed this change or pushed it up.

terraform/data/bq_views/events/pull_request_events.sql
```diff
@@ -0,0 +1,70 @@
...
+FROM (
+  SELECT
+    received,
+    event,
+    PARSE_JSON(payload) AS json_payload
+  FROM
+    `${dataset_id}.${table_id}`)
+WHERE
+  event = "pull_request";
+
```
verbanicm, line 70: nit: newline at end of file

terraform/data/bq_views/events/pull_request_events.sql
```diff
@@ -0,0 +1,70 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- Filters events to those just pertaining to pull requests and parses relevant
+-- pull request fields into columns.
+--
+-- Relevaant GitHub Documentation:
```
verbanicm, line 18:
```suggestion
-- Relevant GitHub Documentation:
```

terraform/data/bq_views/events/pull_request_events.sql
lock14: do we not want newlines at the end of the file?

terraform/bigquery.tf
lock14: hmm strange I tried amending the branch and force pushing, but it didn't appear to take. I just made a separate commit instead.

terraform/data/bq_views/events/pull_request_events.sql
lock14: fixed.

terraform/data/bq_views/events/pull_request_events.sql
verbanicm: There should be, github was flagging it as missing. 🤷🏻‍♂️

terraform/data/bq_views/events/pull_request_events.sql
lock14: How about I just have the `unique_events.sql` view parse the json. Then every view derived from it doesn't need to worry about parsing the json. I think it is valuable having the json parsed as then the full set of JSON fucntions big query provides can be used when creating views/metrics.

lock14: actually my suggestion doesn't work as the unique events view groups by the payload, and JSON values are not allowed in group by expressions apparently.

verbanicm: I tried re-writing this with a CTE and its actually less performant, both speed and data size. I would keep this as is, update to use SAFE_CAST as needed for data types as mentioned.

sethvargo (dismissed):

terraform/data/bq_views/events/pull_request_events.sql
lock14: Mike's suggestion was more performant, So I've updated to use `SAFE_CAST` where needed as per his suggestion.

verbanicm (approved):

PR #112, opened 2023-07-07
Add push events view

lock14:
* Pushes do not contain an underlying resource, thus there is no corresponding resource view
* Ignore terraform directory on pushes and pull requests as it is not needed for the build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,13 @@ on:
   push:
     branches:
       - 'main'
+    paths-ignore:
+      - 'terraform/**'
   pull_request:
     branches:
       - 'main'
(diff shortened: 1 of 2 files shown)


shan4993 (approved):

.github/workflows/ci.yml
```diff
@@ -18,9 +18,13 @@ on:
   push:
     branches:
       - 'main'
+    paths-ignore:
+      - 'terraform/**'
```
verbanicm, line 22: I am hoping that the workflow is skipped here so the required status checks still allow PRs to be merged. I am not sure that will be the case, but wanted to call it out just in  case.

verbanicm (approved):

PR #113, opened 2023-07-07
Add views for Deployment Status data

lock14:
* Added `deployment_status_events.sql` which creates a view that filters events to those only related to Deployment Statuses.
* Added `deployment_statuses.sql` which provides a view of distinct deployment statuses

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,13 +18,9 @@ on:
   push:
     branches:
       - 'main'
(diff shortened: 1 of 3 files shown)


verbanicm:
Ah, well have to remove the path-ignore from the ci.yaml. Was worried this might be the case, but now we know.

verbanicm (dismissed):

verbanicm (approved):

PR #127, opened 2023-09-11
notes

pdewilde:


pipeline/leech/main.go
```diff
@@ -44,6 +44,7 @@ func init() {
 	register.DoFn2x1[context.Context, leech.EventRecord, leech.LeechRecord](&leech.IngestLogsFn{})
 }
 
+// why is this file here and not in cmd/? is it because its not intended to be run from cli?
```
bradegler, line 47: Apache Beam pipelines aren't really executable in the same way as the rest of our commands. Putting it in `cmd` comes with an expectation that it is runnable without a beam runner as is which it is not.

pipeline/leech/main.go
```diff
@@ -112,6 +113,8 @@ func realMain(ctx context.Context) error {
 	}
 
 	var event leech.EventRecord
+	// I know its not really the threat model, but it does seem to be a bad habit to use
+	// string replacement for query building.
```
bradegler, line 117: The go bigquery connector for Apache Beam doesn't support bind parameters so there isn't much choice unless we write our own BQ integration.

pkg/leech/ingest_logs.go
```diff
@@ -124,10 +124,12 @@ func (f *IngestLogsFn) StartBundle(ctx context.Context) error {
 		return fmt.Errorf("failed to read private key: %w", err)
 	}
 	ghAppConfig := githubapp.NewConfig(f.GitHubAppID, f.GitHubInstallID, pk)
+	// Why not f.ghApp = githubapp.New(ghAppConfig)?
```
bradegler, line 127: Initial implementation didn't assign it to f.ghApp and when that change was made it wasn't cleaned up.

pkg/leech/ingest_logs.go
```diff
@@ -124,10 +124,12 @@ func (f *IngestLogsFn) StartBundle(ctx context.Context) error {
...
 	ghApp := githubapp.New(ghAppConfig)
 	f.ghApp = ghApp
 
 	// setup the http client
+	// Why such a massive timeout? Just for large log files?
```
bradegler, line 132: Correct.

pkg/leech/ingest_logs.go
```diff
@@ -138,6 +140,7 @@ func (f *IngestLogsFn) StartBundle(ctx context.Context) error {
 func (f *IngestLogsFn) ProcessElement(ctx context.Context, event EventRecord) LeechRecord {
 	logger := logging.FromContext(ctx)
 
+	// Why have this log as well as the processing element log a few lines down?
```
bradegler, line 143: It could be removed. Its a left over from early debugging.

pkg/leech/ingest_logs.go
```diff
@@ -150,12 +153,15 @@ func (f *IngestLogsFn) ProcessElement(ctx context.Context, event EventRecord) Le
 		RepositoryName:   event.RepositoryName,
 		RepositorySlug:   event.RepositorySlug,
 		LogsURI:          gcsPath,
-		Status:           "SUCCESS",
+		Status:           "SUCCESS", // maybe initialize as PENDING and then update in/after handleMessage?
```
bradegler, line 156: In the majority of cases this should be `SUCCESS` but you certainly could do something else here.

pkg/leech/ingest_logs.go
```diff
@@ -150,12 +153,15 @@ func (f *IngestLogsFn) ProcessElement(ctx context.Context, event EventRecord) Le
...
 	}
+	// Shouldn't this be Infow?
+	// it seems strange to log "result" when we mutate result.Status later on.
```
bradegler, line 159: Fair point, feel free to remove status.

pkg/leech/storage.go
```diff
@@ -61,6 +61,7 @@ func (s *ObjectStore) Write(ctx context.Context, content io.Reader, objectDescri
 	writer := obj.NewWriter(ctx)
 
 	if _, err := io.Copy(writer, content); err != nil {
+		// Should we cancel context? obj.NewWriter() says to do so if you don't want ot write
```
bradegler, line 64: I don't see that in the docs, can you point me to a reference?

pdewilde: https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.NewWriter

>It is the caller's responsibility to call Close when writing is done. To stop writing without saving the data, cancel the context.

pdewilde: Disappointingly, the example code neither closes nor cancels the context.

sethvargo:
@pdewilde do we still need this PR? It looks like it was used as a way to ask questions about the code and we do not intend to merge it.

pdewilde:
> @pdewilde do we still need this PR? It looks like it was used as a way to ask questions about the code and we do not intend to merge it.

Correct, I'll close it out

PR #136, opened 2023-09-26
big query optimization: new table that is partitioned and clustered

pdewilde:
This should make a new BigQuery table that is also subscribed to the existing topic, creating a dual write scenario.

The new table also uses JSON as the type for the payload. It can be fed from a string in BigQuery so it should "just work"


drevell (dismissed):
LGTM for Go readability, I didn't look at the other stuff, others will be more qualified for that

terraform/variables.tf
```diff
@@ -260,6 +266,12 @@ variable "bigquery_project_id" {
   type        = string
 }
 
+variable "bigquery_events_partition_granularity" {
+  description = "How granular you want partition to be. Ideally partitions will be 1-10 GB. Our org had ~750 MB/month. Valid values are DAY, HOUR, MONTH, and YEAR."
+  type        = string
+  default     = "MONTH"
```
verbanicm, line 272: Can you add validation here?

Something like:

```
  validation {
    condition     = contains(["DAY", "HOUR", "MONTH", "YEAR"], var.bigquery_events_partition_granularity)
    error_message = "ERROR: bigquery_events_partition_granularity must be one of DAY, HOUR, MONTH, and YEAR"
  }

```

terraform/bigquery.tf
```diff
@@ -362,6 +461,61 @@ resource "google_bigquery_table" "unique_events_view" {
...
+      JSON_VALUE(payload, "$.repository.name") repository,
+      JSON_VALUE(payload, "$.repository.visibility") repository_visibility,
+      JSON_VALUE(payload, "$.sender.login") sender,
+      SAFE_CAST(JSON_QUERY(payload, "$.sender.id") AS INT64) sender_id,
+    FROM
+      `${google_bigquery_dataset.default.dataset_id}.${google_bigquery_table.raw_events_table.table_id}`
+    WHERE
+      received >= start
+      AND received <= end
+      AND eventTypeFilter == event
```
verbanicm, line 488: nit: `      AND event == eventTypeFilter`

verbanicm, line 488: Is `==` valid?

terraform/bigquery.tf
```diff
@@ -362,6 +461,61 @@ resource "google_bigquery_table" "unique_events_view" {
   ]
 }
 
+# Unique Events -deduplicate rows but as a table function
+resource "google_bigquery_routine" "unique_events_by_date_type" {
```
verbanicm, line 465: Does this need `routine_type`, `language` and/or `return_table_type`?

terraform/pubsub.tf
```diff
@@ -245,6 +245,28 @@ resource "google_pubsub_subscription" "default" {
   }
 }
 
+resource "google_pubsub_subscription" "json" {
```
verbanicm, line 248: Im confused how this is different from the current default subscription?

terraform/bigquery.tf
```diff
@@ -180,6 +180,105 @@ resource "google_bigquery_table_iam_member" "event_viewers" {
   member     = each.value
 }
 
+# Events table with clustering and partitioning / IAM
+resource "google_bigquery_table" "raw_events_table" {
```
verbanicm, line 184: this is the same except adding partition and cluster, why not update the existing one?

terraform/pubsub.tf
pdewilde: Its not. I want two identical subscriptions to the same topic. I plan to dual-write and then later turn off the old stream once everything looks good.

terraform/bigquery.tf
pdewilde: You cannot add partitioning to an existing table

terraform/bigquery.tf
pdewilde: Good questions:

routineType: Probably shout set it

language: defaults to SQL unless you set remoteFunctionOption, so it should be fine, but I could add it

return_table_type: I left absent, though I think either way is fine. If absent, type is inferred at runtime, otherwise wit will cast to match the column types at query time. The former I think is less error prone, as any casting is avoided at the TVF level

terraform/bigquery.tf
pdewilde: Also I changed the schema, events is now JSON rather than STRING

verbanicm: Where? This is the schema for the pubsub topic: https://github.com/abcxyz/github-metrics-aggregator/pull/136/files#diff-c0e7aafb097169e05cfc80fcfa7a7379e723829aebeda7fb0c1113634a8ab785R144-R153

terraform/bigquery.tf
verbanicm: Ack, just wanted to clarify the TF docs didnt specify what the defaults were. Maybe add or document.

terraform/pubsub.tf
verbanicm: Ah knowing that you cant just add a partition, this makes sense now, thanks!!

terraform/bigquery.tf
verbanicm: ....I believe this is what causes the issue when I last tried this.

verbanicm: I guess, have you tested this or are you doing this to test it?

pdewilde: https://cloud.google.com/pubsub/docs/bigquery#protobuf-to-zetasql

So string is the proper protobuf type to import to a JSON column. 

I haven't tested it (figure this will tell us if it will work), but we use the same pattern in PMAP: https://github.com/abcxyz/pmap/blob/main/pkg/server/pubsub.go#L64  https://github.com/abcxyz/pmap/blob/main/terraform/modules/pubsub-bigquery/main.tf#L69

terraform/bigquery.tf
pdewilde: Yeah, the TF documentation was a bit lacking, but the link to the API documentation did show what these do

terraform/bigquery.tf
pdewilde: You're right, should be =

I hate that SQL is different from every other language in the regard

terraform/bigquery.tf
verbanicm: They are using a non-schema approach that has a specific format from pubsub, so I dont think this will work but happy to test it out.

[Details part 1](https://cloud.google.com/pubsub/docs/subscription-properties#bigquery:~:text=If%20you%20don%27t%20select%20the%20Use%20topic%20schema%20option%2C%20ensure%20that%20the%20BigQuery%20table%20has%20a%20column%20called%20data%20of%20type%20BYTES%2C%20STRING%2C%20or%20JSON.%20Pub/Sub%20writes%20the%20message%20to%20this%20BigQuery%20column.)
[Details part 2](https://cloud.google.com/pubsub/docs/subscription-properties#write-metadata)

pdewilde: I see, it is a bit different. Only one way to find out if it works

terraform/variables.tf
pdewilde: Thanks for the suggestion, done

pkg/webhook/bigquery.go
```diff
@@ -111,6 +111,7 @@ func (bq *BigQuery) WriteFailureEvent(ctx context.Context, failureEventTableID,
 	return nil
 }
 
+// TODO: limit by time period to avoid unnecessary scanning -- low priority
```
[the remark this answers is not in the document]
pdewilde: done

terraform/bigquery.tf
pdewilde: I added language and routineType explicitly, and left out return_table_type as its already well documented in the tf docs

terraform/bigquery.tf
```diff
@@ -180,6 +180,105 @@ resource "google_bigquery_table_iam_member" "event_viewers" {
...
+  project = data.google_project.default.project_id
+
+  deletion_protection = true
+  table_id            = var.raw_events_table_id
+  dataset_id          = google_bigquery_dataset.default.dataset_id
+  schema = jsonencode([
+    {
+      "name" : "delivery_id",
+      "type" : "STRING",
+      "mode" : "NULLABLE",
```
stytchiz, line 194: NULLABLE is the default so this apparently can be omitted. However, do we expect NULL values here? https://cloud.google.com/bigquery/docs/schemas#modes

terraform/bigquery.tf
```diff
@@ -180,6 +180,105 @@ resource "google_bigquery_table_iam_member" "event_viewers" {
...
+    {
+      "name" : "payload",
+      "type" : "JSON",
+      "mode" : "NULLABLE",
+      "description" : "Event payload JSON"
+    }
+  ])
+
+  time_partitioning {
+    field = "received" # TODO: would we rather extract an actual time from the payload?
```
stytchiz, line 224: I don't really understand this in the context of TODO.  Can you rephrase it to something more imperative?

http://go/todo-style

terraform/bigquery.tf
```diff
@@ -362,6 +461,63 @@ resource "google_bigquery_table" "unique_events_view" {
...
+      JSON_VALUE(payload, "$.repository.visibility") repository_visibility,
+      JSON_VALUE(payload, "$.sender.login") sender,
+      SAFE_CAST(JSON_QUERY(payload, "$.sender.id") AS INT64) sender_id,
+    FROM
+      `${google_bigquery_dataset.default.dataset_id}.${google_bigquery_table.raw_events_table.table_id}`
+    WHERE
+      received >= start
+      AND received <= end
+      AND event = eventTypeFilter
+    GROUP BY
```
stytchiz, line 491: Does it work to use SELECT DISTINCT and remove this huge GROUP BY clause?

https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_distinct

stytchiz (approved):

terraform/bigquery.tf
pdewilde: Hopefully not, but I think the idea is to take whatever GitHub gives us, which is why they were all set to null able

terraform/bigquery.tf
pdewilde: That was a note to myself I forgot to remove. The received time is when GMA gets the webhook from GH, not when the event was actually created. Ideally, we would parse that value from the payload and use it instead.

I think I'll put that off for now and make an enhancement issue in the GH repo.

terraform/bigquery.tf
pdewilde: Hmm, that may be possible, I'll look into it.

terraform/bigquery.tf
pdewilde: https://github.com/abcxyz/github-metrics-aggregator/issues/139

terraform/bigquery.tf
pdewilde: I was copying the existing format from the other table as far as including the optional field.

I think its useful, most people reading this are probably not BigQuery experts and this makes it immediately apparent that null values are permissible without having to look at the documentation.

terraform/bigquery.tf
pdewilde: I'll have to experiment with that

pdewilde: Though as @lock14 pointed out in a chat thread, there are duplicates which we are not catching due to installation id being included in retried events that actually were successfully written, as payloads vary slightly.

I think that duplicate events may be better handled with a scheduled query than an TVF.

pdewilde: I did update the existing view and this TVF to use SELECT_DISTINCT as in testing it was identical, thanks for the suggestion!

verbanicm (approved):

drevell (approved):

terraform/pubsub.tf
```diff
@@ -245,6 +245,28 @@ resource "google_pubsub_subscription" "default" {
...
+  project = var.project_id
+
+  name  = "${var.prefix_name}-bq-json-sub"
+  topic = google_pubsub_topic.default.name
+
+  bigquery_config {
+    table            = format("${google_bigquery_table.raw_events_table.project}:${google_bigquery_table.raw_events_table.dataset_id}.${google_bigquery_table.raw_events_table.id}")
```
capri-xiyue, line 255: I think this should be "${google_bigquery_table.raw_events_table.project}.${google_bigquery_table.raw_events_table.dataset_id}.${google_bigquery_table.raw_events_tabletable_id}", see example here https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/pubsub_subscription.html#example-usage---pubsub-subscription-push-bq. @pdewilde

pdewilde: Thank you, that was a really good catch!

PR #142, opened 2023-09-28
Use Large Runners in Build Jobs

pdewilde:
The existing CI build job took over 5 minutes to run. This will use a new runner group locked down to the github-metrics-aggregator repo called "beefy" which has a 4-core and a 8-core runner to help speed up ci jobs significantly.


sailorlqh (approved):

.github/workflows/ci.yml
```diff
@@ -79,7 +79,8 @@ jobs:
 
   # Build github-metrics-aggregator image and push to artifact registry
   build:
-    runs-on: 'ubuntu-latest'
+    runs-on:
```
sethvargo, line 82: Can you add a comment here with some background context?

sethvargo (dismissed):

.github/workflows/ci.yml
pdewilde: I figured I wouldn't actually merge this, as pdewilde-test is probably not the final label we will end up with

But unfortunatly it doesn't seem to be working at all

pdewilde:
image

Seems like 8 or 16 cores probably is the sweet spot. Billing is per-minute, so currently 16 cores isn't as bad as it looks, as it finishes in under 2 minutes, while 8 cores will be billed for the full 3 minutes. Of course if build times increase this could flip.

sethvargo:
@pdewilde I created a new runner group (Beefy) with two machine configurations (4-core and 8-core).

sailorlqh (approved):

.github/workflows/ci.yml
```diff
@@ -79,7 +79,8 @@ jobs:
...
+      group: 'Beefy' # group with a 4-core and 8-core runner for faster builds
```
sethvargo, line 83: @pdewilde - will it randomly choose one? I created the _group_ because that's the only way to restrict the runners to a specific repo. I was thinking you would use either the 4-core or 8-core machine label directly.

pdewilde: Aah, I assumed you intended for me to use the group. I'm not sure what the logic is. I did one test and saw it corresponded to the 8-core machine and was hoping it would use the more powerful machine first.

I'll have to take a look

pdewilde: https://github.com/actions/runner/issues/1665

Looks like there is a long-existing ask but not movement on such a prioritization feature in a year. (That is for Self-Hosted rather than Large runners, but I would imagine the behavior is similar).

Probably best to just slap on the 8-core machine label directly. Probably no need to have a 4-core for now in the Beefy group as most of the other jobs run in pretty similar times.

PR #147, opened 2023-10-09
Add Beam DoFn to process a commit

lock14:
Added an Apache Beam DoFn to process a commit and determine its review status in GitHub.

### DIFF
--- a/pkg/review/commit_review_status.go
+++ b/pkg/review/commit_review_status.go
@@ -21,13 +21,19 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/abcxyz/pkg/logging"
 	"github.com/shurcooL/githubv4"
 	"golang.org/x/oauth2"
 )
 
-// githubPRApproved is the value GitHub set the `reviewDecision` field to
-// when a Pull Request has been approved by a reviewer.
-const githubPRApproved = "APPROVED"
+const (
+	// githubPRApproved is the value GitHub set the `reviewDecision` field to
+	// when a Pull Request has been approved by a reviewer.
+	githubPRApproved = "APPROVED"
+
+	// the default approval status we assign to a commit.
+	defaultApprovalStatus = "UNKNOWN"
+)
 
 // commitQuery is the BigQuery query that selects the commits that need
 // to be processed. The criteria for a commit that needs to be processed are:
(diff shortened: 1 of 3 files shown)


pkg/review/commit_review_status.go
```diff
@@ -29,6 +30,9 @@ import (
 // when a Pull Request has been approved by a reviewer.
 const githubPRApproved = "APPROVED"
 
+// the default approval status we assign to a commit.
```
drevell, line 33: Optional: you can combine these into a const block:

```
const (
  githubPRApproved = ...
  defaultApprovalStatus = ...
)

File: pkg/review/commit_review_status.go [hunk 2]
```diff
@@ -91,6 +106,71 @@ type PullRequest struct {
...
+
+// ProcessElement is a DoFn implementation that take a Commit, determines
+// if the commit was properly approved, and outputs the resulting
+// CommitReviewStatus using the provided emit function.
+// A commit is considered properly reviewed as long as there is an associated
+// PR for the commit targeting the repository's main branch with reviewDecision
+// of 'APPROVED'.
+func (fn *CommitApprovalDoFn) ProcessElement(ctx context.Context, commit Commit, emit func(CommitReviewStatus)) {
+	logger := logging.FromContext(ctx)
+	logger.Infow("processing commit", "commit", commit)
```
drevell, line 144: I think we prefer InfoContext over Info. https://github.com/abcxyz/readability/wiki/Go-Best-Practices#logging

Same throughout this file; FooContext() vs Foo()

pkg/review/commit_review_status.go
```diff
@@ -91,6 +106,71 @@ type PullRequest struct {
...
+	requests, err := GetPullRequestsTargetingDefaultBranch(ctx, fn.GithubClient, commit.Organization, commit.Repository, commit.SHA)
+	if err != nil {
+		logger.Errorf("failed to get pull requests for commit: %v", err)
+		return // this commit could not be processed
```
drevell, line 148: It's a little concerning that this error isn't handled in any way besides logging

drevell (changes requested):

pkg/review/commit_review_status.go
lock14: the version of abcxyz/pkg that was being used was outdated. I've opened a PR to update it: #148. Once that's merged I will rebase this branch and updatre these logging statements.

pkg/review/commit_review_status.go
lock14: There are essentially two different kind of errors that could happen:
1. **Transient Errors**: We aren't able to get the pull requests for a commit because of some temporary issue with GitHub (e.g. GitHub servers are down). In these cases the commit will simply be retried the next time the pipeline is run.
2. **Perment Errors**: There is something wrong with the commit data itself that makes GitHub return an error. For example, the repository the commit came from may have been deleted so we no longer can get pull request information from GitHub about it.

For the **Transient Errors**, the commit will be retried during the next pipleine execution so there is not need to do anything else aside from logging the error. For **Perment Errors** it may be useful to do something aside from logging, but it is hard to say exactly what should be done without seeing what kinds of errors like this occur and how frequently. My thought was to log the error for now and then consider more sophisticated error handling if we need it.

pkg/review/commit_review_status.go
lock14: Good suggestion, done.

pkg/review/commit_review_status.go
lock14: This has been fixed.

pkg/review/commit_review_status.go
drevell: Could you add this as a code comment?

lock14: done.

drevell (approved):

PR #148, opened 2023-10-10
Upgrade github.com/abcxyz/pkg to v0.7.0

lock14:


.github/workflows/ci.yml
```diff
@@ -61,263 +62,263 @@ jobs:
   go_lint:
     uses: 'abcxyz/pkg/.github/workflows/go-lint.yml@main' # ratchet:exclude
     with:
-      go_version: '1.20'
+      go_version: '1.21'
```
drevell, line 65: Should these use `${{env.go_version}}` rather than hardcoding 1.21 (throughout this file)?

drevell (dismissed):

.github/workflows/ci.yml
verbanicm, line 1: Can you format this again? There should be 2 spaces per indent, not sure what happened here....

lock14: done

.github/workflows/ci.yml
verbanicm: I believe capitalization matters, the env is UPPER and you gave lower, maybe try that?

drevell (approved):

PR #149, opened 2023-10-11
Refactor to use bigqueryio.QualifiedTableName

lock14:
* Change references to BigQuery table names to use the `bigqueryio.QualifiedTableName` struct. This allows the underlying tables to vary gcp project and big query dataset independently from each other while also provided a cleaner way to refer to a fully qualfied bigquery table.

### DIFF
--- a/pkg/review/commit_review_status.go
+++ b/pkg/review/commit_review_status.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	"github.com/abcxyz/pkg/logging"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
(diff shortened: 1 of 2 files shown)


pkg/review/commit_review_status.go
```diff
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	"github.com/abcxyz/pkg/logging"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
```
drevell, line 25: Optional style choice: save a few characters by renaming this import:

```
bqio "github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
```

drevell (approved):

PR #151, opened 2023-10-16
Add Beam DoFn to lookup break glass issue

lock14:
Added an Apache Beam DoFn to process a `CommitReviewStatus` struct and populate its `breakGlassIssueURLs` field with the associated break glass isssue(s) (if any are present).

### DIFF
--- a/pkg/review/bq/bq.go
+++ b/pkg/review/bq/bq.go
@@ -1,0 +1,55 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package bq contains abstracts away the boilerplate code that results from
+// mapping BigQuery table rows to and from a struct
+
+package bq
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"cloud.google.com/go/bigquery"
+	"google.golang.org/api/iterator"
+)
+
+// Query takes a queryString (assumed to be valid SQL) and executes it against
+// BigQuery using the given client. The results are then mapped to a slice of T,
+// where each row in the result is mapped to a struct of type T.
+func Query[T any](ctx context.Context, client *bigquery.Client, queryString string) ([]T, error) {
+	query := client.Query(queryString)
+	job, err := query.Run(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("query.Run failed: %w", err)
+	}
+	rows, err := job.Read(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("job.Read failed: %w", err)
+	}
+	items := make([]T, 0, rows.TotalRows)
+	var t T
+	for {
+		err := rows.Next(&t)
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("rows.Next failed: %w", err)
+		}
+		items = append(items, t)
+	}
+	return items, nil
+}
--- a/pkg/review/commit_review_status.go
+++ b/pkg/review/commit_review_status.go
@@ -21,6 +21,8 @@ import (
 	"context"
 	"fmt"
 
+	"cloud.google.com/go/bigquery"
+	"github.com/abcxyz/github-metrics-aggregator/pkg/review/bq"
 	"github.com/abcxyz/pkg/logging"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
 	"github.com/shurcooL/githubv4"
@@ -28,12 +30,12 @@ import (
 )
 
 const (
-	// githubPRApproved is the value GitHub set the `reviewDecision` field to
+	// GithubPRApproved is the value GitHub set the `reviewDecision` field to
 	// when a Pull Request has been approved by a reviewer.
-	githubPRApproved = "APPROVED"
+	GithubPRApproved = "APPROVED"
 
 	// the default approval status we assign to a commit.
-	defaultApprovalStatus = "UNKNOWN"
+	DefaultApprovalStatus = "UNKNOWN"
 )
 
 // commitQuery is the BigQuery query that selects the commits that need
@@ -90,10 +92,16 @@ type Commit struct {
 // BigQuery.
 type CommitReviewStatus struct {
 	Commit
-	HTMLURL        string `bigquery:"commit_html_url"`
-	PullRequestID  int    `bigquery:"pull_request_id"`
-	ApprovalStatus string `bigquery:"approval_status"`
-	BreakGlassURL  string `bigquery:"break_glass_issue_url"`
+	HTMLURL        string   `bigquery:"commit_html_url"`
+	PullRequestID  int      `bigquery:"pull_request_id"`
+	ApprovalStatus string   `bigquery:"approval_status"`
+	BreakGlassURLs []string `bigquery:"break_glass_issue_urls"`
+}
+
+// breakGlassIssue is a struct that maps the columns of the result of
+// the breakGlassIssueQuery.
+type breakGlassIssue struct {
+	HTMLURL string `biquery:"html_url"`
 }
 
 // PullRequest represents a pull request in GitHub and contains the
@@ -107,6 +115,31 @@ type PullRequest struct {
 	DatabaseID     githubv4.Int
 	Number         githubv4.Int
 	ReviewDecision githubv4.String
+}
+
+// BreakGlassIssueFetcher fetches break glass issues from a data source.
+type BreakGlassIssueFetcher interface {
+	// getBreakGlassIssues retrieves all break glass issues created by the given
+	// author and whose open duration contains the specified timestamp.
+	// The issue's open duration contains the timestamp if
+	// issue.created_at <= timestamp <= issue.closed_at holds.
+	getBreakGlassIssues(ctx context.Context, author, timestamp string) ([]breakGlassIssue, error)
+}
+
+// BigQueryBreakGlassIssueFetcher implements the BreakGlassIssueFetcher
+// interface and fetches the break glass issue data from BigQuery.
+type BigQueryBreakGlassIssueFetcher struct {
+	client *bigquery.Client
+	config *CommitApprovalPipelineConfig
+}
+
+func (bqif *BigQueryBreakGlassIssueFetcher) getBreakGlassIssues(ctx context.Context, author, timestamp string) ([]breakGlassIssue, error) {
+	issueQuery := GetBreakGlassIssueQuery(bqif.config.IssuesTable, author, timestamp)
+	items, err := bq.Query[breakGlassIssue](ctx, bqif.client, issueQuery)
+	if err != nil {
+		return nil, fmt.Errorf("client.Query failed: %w", err)
+	}
+	return items, nil
 }
 
 // CommitApprovalPipelineConfig holds the configuration data for the
@@ -133,6 +166,14 @@ type CommitApprovalPipelineConfig struct {
 type CommitApprovalDoFn struct {
 	Config       CommitApprovalPipelineConfig
 	GithubClient *githubv4.Client
+}
+
+// BreakGlassIssueDoFn is an object that implements beams "DoFn" interface to
+// provide the processing logic for converting retrieving the associated break
+// glass issue for a CommitReviewStatus.
+type BreakGlassIssueDoFn struct {
+	Config                 CommitApprovalPipelineConfig
+	BreakGlassIssueFetcher BreakGlassIssueFetcher
 }
 
 // ProcessElement is a DoFn implementation that take a Commit, determines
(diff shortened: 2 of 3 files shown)


pkg/review/commit_review_status.go
```diff
@@ -135,6 +142,14 @@ type CommitApprovalDoFn struct {
 	GithubClient *githubv4.Client
 }
 
+// BreakGlassIssueDoFn is an object that implements beams "DoFn" interface to
+// provide the processing logic for converting retrieving the associated break
+// glass issue for a CommitReviewStatus.
+type BreakGlassIssueDoFn struct {
```
sethvargo, line 148: It's more common to name function types with the `Func` suffix (e.g. `context.CancelFunc`). However, this appears to be a struct.

pkg/review/bq/bq.go
```diff
@@ -0,0 +1,70 @@
...
+	"errors"
+	"fmt"
+
+	"cloud.google.com/go/bigquery"
+	"google.golang.org/api/iterator"
+)
+
+// Client represents a bigquery client that automatically maps table
+// rows to and from structs of type T.
+type Client[T any] interface {
```
sethvargo, line 31: What is the goal with this interface and what is the goal with making it generic? We seem to just be wrapping the native bigquery interface.

pkg/review/commit_review_status.go
```diff
@@ -192,12 +207,59 @@ func (fn *CommitApprovalDoFn) ProcessElement(ctx context.Context, commit Commit,
...
+			// We should only get transient style errors from BigQuery
+			// (e.g. network is down etc.). So, we can just log the error and then
+			// drop this CommitReviewStatus from the pipeline. It will then get
+			// retried on the next run of the pipeline.
+			logger.ErrorContext(ctx, "failure when trying to get break glass issue: %v", err)
+			return
+		}
+		commitReviewStatus.BreakGlassURL = breakGlassIssue.HTMLURL
+	}
+	emit(commitReviewStatus)
```
sethvargo, line 238: We should check if `emit` is not nil before calling.

pkg/review/commit_review_status.go
```diff
@@ -192,12 +207,59 @@ func (fn *CommitApprovalDoFn) ProcessElement(ctx context.Context, commit Commit,
...
+}
+
+// getBreakGlassIssue Searches BigQuery for a break glass issue that was opened
+// by the given author and whose created_at date was on or before the given
+// timestamp and whose closed_at date was on or after the given timestamp.
+func getBreakGlassIssue(ctx context.Context, client bq.Client[breakGlassIssue], issuesTable bigqueryio.QualifiedTableName, author, timestamp string) (*breakGlassIssue, error) {
```
sethvargo, line 244: Why wouldn't this just be a function on our client interface instead?

pkg/review/commit_review_status.go
```diff
@@ -192,12 +207,59 @@ func (fn *CommitApprovalDoFn) ProcessElement(ctx context.Context, commit Commit,
...
+	issueQuery := GetBreakGlassIssueQuery(issuesTable, author, timestamp)
+	items, err := client.Query(ctx, issueQuery)
+	if err != nil {
+		return nil, fmt.Errorf("client.Query failed: %w", err)
+	}
+	var breakGlassIssue *breakGlassIssue
+	if len(items) > 0 {
+		breakGlassIssue = &items[0]
+	}
```
sethvargo, line 253: Should we handle a situation where there's more than one result? Would that be an error or a "should never be possible"? If there can be multiple results, are you sure we want the first one and not the last? Can we guarantee the results are returned in the same order?

pkg/review/bq/bq.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("query.Run failed: %w", err)
+	}
+	rows, err := job.Read(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("job.Read failed: %w", err)
+	}
+	items := make([]T, 0, rows.TotalRows)
+	var t T
+	for err := rows.Next(&t); !errors.Is(err, iterator.Done); err = rows.Next(&t) {
```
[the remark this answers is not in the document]
lock14: This is a pretty common for loop pattern for working with iterators, though its probably more prevalent in C/C++ worlds. Personally, I find an infinite loop with a break somewhere in the middle of the loop harder to follow as it tends to obfuscate the loops terminal condition. I'll update it to the suggested style though if that the pattern we want to use.

pkg/review/commit_review_status.go
lock14: This struct is an Apache Beam DoFn style struct (https://beam.apache.org/documentation/programming-guide/#4212-creating-a-dofn) which is what the name is intending to convey.

pkg/review/bq/bq.go
lock14: I needed to have some indirection so I can test my business logic without actually making calls to BigQuery. 
See https://github.com/abcxyz/github-metrics-aggregator/blob/ce6de05391ff0b2d1a05905a987b710f2faf5dbc/pkg/review/commit_review_status_test.go#L1278
For how this is used in the tests.

As for making it generic, the client logic works regardless of struct type so I figured why not make it generic, especially since it seems like a common pattern to want ton take each row from a a BigQuery result and map it to a struct, so having a single piece of code to let you do that seemed like a useful thing.

pkg/review/commit_review_status.go
lock14: There should really only ever be at most one result returned. While I think it may technically possible for someone to open a second break glass issue if they already have one open, I can't think of a reason why someone would even do that as they already would already have break glass permissions from the first break glass issue they opened.

For the pipeline code though, all we really care about is whether the author had break glass permissions or not during the time the commit was made. So if the author had multiple break glass issues open at the same time, then any one of the issues is sufficient to establish that they had break glass permissions. So just using the first one is fine.

pkg/review/commit_review_status.go
lock14: Because the client is generic and this function is is mostly related to my specific business logic of getting the break glass issue. I could refactor the parts pertaining to executing a query that you only expect to have 1 into the client. Something like:
```
func (bqc *bigQueryClientImpl[T]) QueryAndGetFirst(ctx context.Context, queryString string) (*T, error) {
	items, err := bqc.Query(ctx, queryString)
	if err != nil {
		return nil, fmt.Errorf("bqc.Query failed: %w", err)
	}
	var t *T
	if len(items) > 0 {
		t = &items[0]
	}
	return t, nil
}
```
Then this function could be rewritten as:
```
func getBreakGlassIssue(ctx context.Context, client bq.Client[breakGlassIssue], issuesTable bigqueryio.QualifiedTableName, author, timestamp string) (*breakGlassIssue, error) {
	issueQuery := GetBreakGlassIssueQuery(issuesTable, author, timestamp)
	breakGlassIssue, err := client.QueryAndGetFirst(ctx, issueQuery)
	if err != nil {
		return nil, fmt.Errorf("client.QueryAndGetFirst failed: %w", err)
	}
	return breakGlassIssue, nil
}
```
I'm not sure if this is significantly better or not, but you can let me know what you think.

pkg/review/commit_review_status.go
lock14: The `emit` function supplied by the Apache Beam runner and will never be nil, so I don't think this is necessary.

pkg/review/bq/bq.go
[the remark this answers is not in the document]
lock14: I've updated the for loop to the suggested style.

pkg/review/commit_review_status.go
sethvargo: Ah TIL. Thanks!

pkg/review/bq/bq.go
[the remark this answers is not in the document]
sethvargo: Thanks!

pkg/review/bq/bq.go
sethvargo: I think we might be conflating a few different problems in to the same solution. 

For testing, an interface is the correct answer. However, the interface should be _specific_ like `GetBreakGlassIssue`, not generic like `query`. For example:

```go
type BreakGlassIssuer interface {
  BreakGlassIssue(ctx context.Context, author, timestamp string) (*BreakglassIssue, error)
}
```

Then you can have multiple implementations of that interface:

```go
# real
type BigQueryBreakGlassIssuer struct {
  client *bigquery.Client
  config *Config
}

func NewBigQueryBreakGlassIssuer() (*BigQueryBreakGlassIssuer, error) {
  // TODO
}

type (b *BigQueryBreakGlassIssuer) BreakGlassIssue(ctx context.Context, author, timestamp string) (*BreakglassIssue, error) {
   // basically the same logic as L244 
}
```

```go
# test
type TestBreakGlassIssuer struct {
  results *BreakGlassIssue
  err error
}

type (b *BigQueryBreakGlassIssuer) BreakGlassIssue(ctx context.Context, author, timestamp string) (*BreakglassIssue, error) {
  return b.results, err
}
```

As for generics, the current implementation requires each _client_ to declare its structure up front; it does not implement a generic generic; you cannot have functions that return a different type that the struct. It seems really expensive to create a client for each data type I'd need to return.

pkg/review/commit_review_status.go
sethvargo: Defensive programming is always necessary :smile:. Without it, one day their API will change and we'll waste hours chasing down a NPE.

pkg/review/commit_review_status.go
sethvargo: Similar to what [I said here](https://github.com/abcxyz/github-metrics-aggregator/pull/151#discussion_r1361413242), I think we're conflating a generic client with an interface that abstracts business logic. I think you want your interface to be at a _higher_ level than anything BigQuery or database-specific.

pkg/review/commit_review_status.go
sethvargo: Hmm - I wonder if we should attach _all_ breakglass issues (slice instead of string) to the PR.

pkg/review/commit_review_status.go
lock14: A `nil` emit function would be such a wrong thing to pass in that I don't think silently ignoring it and effectively just dropping all commits would be preferable. I added a nil check and some logging in the nil case so that if this does ever happen we have some indication that a major assumption of the pipeline has been violated.

pkg/review/bq/bq.go
lock14: Yeah this is a limitation of Go's implementation of generics. Ideally you'd put the type parameter on the receiver function itself instead of the client, e.g. something like:
```
type Client interface {
	Query[T any](ctx context.Context, queryString string) ([]T, error)
}
```
but alas it's not possible to do that in Go.

I agree that that if you were working with many different types of structs having a Client each type would be expensive and clunky. But if you were mostly working with a one struct type (maybe your writing CRUD style API or something) then I don't think it would be that unreasonable. 

Personally, I don't think a generic interface has to be reasonable to use in every conceivable instance to justify its existence, just that there is a reusable bit of code that can be reasonably conceived to be reused in more than one instance (In this instance, mapping a big query table to a slice seemed like a common enough use case that making it generic seemed useful). 

You are correct though, for my testing needs all I need is to abstract is the the code to execute fetching the BreakGlassIssue. I'm not super married to this bit of abstraction so I've refactored it to be more along the lines of your suggestions.

pkg/review/commit_review_status.go
lock14: It wouldn't change the pipeline architecture that much to accommodate storing more than one break glass issue if there happen to be more than one. I've updated the code accordingly.

pkg/review/commit_review_status.go
sethvargo: In that case, I'd still expect us to catch the error and raise a specific `panic` instead of the default generic panic that Go emits (which can be very difficult to track down).

lock14: I see, I'm not too familiar with panics. Would that look something like this?
```
if emit == nil {
    panic("nil emit function was provided")
}
```

pkg/review/commit_review_status.go
lock14: I've update the interface to be at a higher level of abstraction. Let me know what you think

pkg/review/bq/bq.go
lock14: removed the client and refactored interface. let me know what you think.

pkg/review/commit_review_status.go
lock14: I've updated the function to panic immediately if a `nil` emit function is passed in.

pkg/review/commit_review_status.go
```diff
@@ -192,12 +233,42 @@ func (fn *CommitApprovalDoFn) ProcessElement(ctx context.Context, commit Commit,
...
+			// drop this CommitReviewStatus from the pipeline. It will then get
+			// retried on the next run of the pipeline.
+			logger.ErrorContext(ctx, "failure when trying to get break glass issue: %v", err)
+			return
+		}
+		commitReviewStatus.BreakGlassURLs = mapSlice(breakGlassIssues, func(issue breakGlassIssue) string {
+			return issue.HTMLURL
+		})
+	}
+	emit(commitReviewStatus)
```
sethvargo, line 263: Can/should this be backgrounded (`go emit(commitReviewStatus)`)? Or do we want to block during the emission?

sethvargo (approved):

pkg/review/commit_review_status.go
lock14: From all the examples int the Apache Beam documentation the `emit` function is directly called and not spun off into a separate goroutine. I couldn't find anything in the docs saying this was a requirement, but I think keeping it as is (a potential blocking call) is fine.

PR #153, opened 2023-10-17
Fix table function, end is a reserved keyword and cannot be used as t…

pdewilde:
…he name of an argument. Switched end to endDate

### DIFF
--- a/terraform/bigquery.tf
+++ b/terraform/bigquery.tf
@@ -473,17 +473,17 @@ resource "google_bigquery_routine" "unique_events_by_date_type" {
     FROM
       `${google_bigquery_dataset.default.dataset_id}.${google_bigquery_table.raw_events_table.table_id}`
     WHERE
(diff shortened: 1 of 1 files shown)


terraform/bigquery.tf
```diff
@@ -473,17 +473,17 @@ resource "google_bigquery_routine" "unique_events_by_date_type" {
     FROM
       `${google_bigquery_dataset.default.dataset_id}.${google_bigquery_table.raw_events_table.table_id}`
     WHERE
-      received >= start
-      AND received <= end
+      received >= startDate
```
sethvargo, line 476: Nit: is this a date or a datetime? If it's a date, +1. If it's a datetime, maybe `startTimestamp`?

pdewilde: Changed

drevell (approved):

PR #154, opened 2023-10-18
Pdewilde/fix query syntax

pdewilde:
For the unique_events_view, JSON_VALUE is a more appropriate function as we would only ever cast a scalar value to an int anyways.

For the table function, payload is now a json type, so different syntax was more appropriate (also existing code was invalid due to differences in the return type of JSON_QUERY for a json string vs a json expression.

### DIFF
--- a/terraform/bigquery.tf
+++ b/terraform/bigquery.tf
@@ -473,17 +473,17 @@ resource "google_bigquery_routine" "unique_events_by_date_type" {
     FROM
       `${google_bigquery_dataset.default.dataset_id}.${google_bigquery_table.raw_events_table.table_id}`
     WHERE
-      received >= start
-      AND received <= end
+      received >= startDate
+      AND received <= endDate
       AND event = eventTypeFilter
     EOT
 
   arguments {
-    name      = "start"
+    name      = "startDate"
     data_type = jsonencode({ typeKind : "TIMESTAMP" })
   }
   arguments {
-    name      = "end"
+    name      = "endDate"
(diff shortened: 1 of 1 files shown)


terraform/bigquery.tf
```diff
@@ -462,14 +462,14 @@ resource "google_bigquery_routine" "unique_events_by_date_type" {
...
       payload,
-      JSON_VALUE(payload, "$.organization.login") organization,
-      SAFE_CAST(JSON_VALUE(payload, "$.organization.id") AS INT64) organization_id,
-      JSON_VALUE(payload, "$.repository.full_name") repository_full_name,
-      SAFE_CAST(JSON_QUERY(payload, "$.repository.id") AS INT64) repository_id,
-      JSON_VALUE(payload, "$.repository.name") repository,
-      JSON_VALUE(payload, "$.repository.visibility") repository_visibility,
-      JSON_VALUE(payload, "$.sender.login") sender,
-      SAFE_CAST(JSON_QUERY(payload, "$.sender.id") AS INT64) sender_id,
+      LAX_STRING(payload.organization.login) organization,
```
stytchiz, line 465: Why not SAFE.STRING(...)?

stytchiz (approved):

terraform/bigquery.tf
pdewilde: To keep the behavior the same to how the existing query treated non-scalar values. This will give null while SAFE.STRING would give an error.

PR #163, opened 2023-10-30
Add Commit Review Status Beam Pipeline

lock14:
* Add function to construct and wire together the different components of the commit review status beam pipeline
* Add main file that compiles down to an executable to be run on gcp dataflow.


pipeline/review/main.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/apache/beam/sdks/v2/go/pkg/beam"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/x/beamx"
+)
+
+var (
+	GithubToken             string
+	PushEventsTable         string
+	CommitReviewStatusTable string
+	IssuesTable             string
```
yolocs, line 39: No need to be public?

pkg/review/commit_review_status.go
```diff
@@ -167,16 +183,51 @@ type CommitApprovalPipelineConfig struct {
 // CommitApprovalDoFn is an object that implements beams "DoFn" interface to
 // provide the processing logic for converting a Commit to CommitReviewStatus.
 type CommitApprovalDoFn struct {
-	Config       CommitApprovalPipelineConfig
-	GithubClient *githubv4.Client
+	Config CommitApprovalPipelineConfig
```
yolocs, line 186: Can this be lower "c" `config`? No need to expose it if not necessary.
Can this be a pointer? `*CommitApprovalPipelineConfig`? More of a convention - use pointer whenever possible

yolocs (dismissed):

pkg/review/commit_review_status.go
lock14: it needs to be public so that it can be serialized. Apache Beam will only serialize public field of a struct. For similar reasons I don't think using a pointer makes sense here as I don't think beam will serialize it correctly if it's a pointer.

pipeline/review/main.go
lock14: changed to private.

pkg/review/commit_review_status.go
yolocs: Curious - Does it require the `Config` being public or the just the fields in it? 

Could you add a short comment above the config about serializing? I know you've documented somewhere else but I think it's better to stay closer to the thing to be serialized.

yolocs (approved):

yolocs (approved):

PR #166, opened 2023-10-31
Fix sneaky bug where TVF sql statements cannot end in ; due to being wrapped in another SQL statement by terraform provider

pdewilde:
TF will actually call something like

```sql
CREATE OR REPLACE TABLE FUNCTION mydataset.name(argument INT64)
AS (
  ${YOUR_QUERY_HERE}
);
```

Unfortunately, this means it looks different to how views are declared, as they seem to tolerate ending in a `;`, and the actual call doesn't show up in plan or apply logs.


gjonathanhong:
I think you're right. The docs hint at, 
```
If language=SQL, it is the substring inside (but excluding) the parentheses.
```

gjonathanhong (approved):

sethvargo (approved):

PR #169, opened 2023-11-01
Fix bug with command line arg validation

lock14:
* Beam requires calling `beam.init()` prior to doing validation on command-line flags.
   https://cloud.google.com/dataflow/docs/guides/setting-pipeline-options#CreatePipelineFromArgs

### DIFF
--- a/pipeline/review/main.go
+++ b/pipeline/review/main.go
@@ -66,13 +66,14 @@ func main() {
 func realMain(ctx context.Context) error {
 	// parse flags
 	flag.Parse()
+	// initialize beam. This must be done after flag.Parse() but before any flag validation.
+	// https://cloud.google.com/dataflow/docs/guides/setting-pipeline-options#CreatePipelineFromArgs
+	beam.Init()
 	// parse commandline arguments into config
 	pipelineConfig, err := getConfigFromFlags()
 	if err != nil {
 		return fmt.Errorf("failed to construct pipeline config: %w", err)
 	}
-	// initialize beam.
-	beam.Init()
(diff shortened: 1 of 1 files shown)


sethvargo:
Wouldn't it be a bug if options are passed then?

lock14:
> Wouldn't it be a bug if options are passed then?

It would be would if say you wanted to run the pipeline locally via manual execution. However, when executing the pipeline via Dataflow in GCP, the Dataflow runner will use the `options` flag when starting up a worker node to pass in the pipeline state, e.g. it will do something like:
```
./review --options=<pipeline-state>
```
I don't know if there is a good way to tell if the binary is being invoked by a Dataflow worker node or not, so I don't think we can do something like fail if the `options` flag is set as that would prevent the worker node from reading the pipeline state.

sethvargo:
> I don't know if there is a good way to tell if the binary is being invoked by a Dataflow worker node or not, so I don't think we can do something like fail if the `options` flag is set as that would prevent the worker node from reading the pipeline state.

We could look for [any of these envvars](https://cloud.google.com/workflows/docs/reference/environment-variables#list_of_built-in_environment_variables).

yolocs:
I wonder how much we can bet on `-options` being a stable flag set by dataflow? Maybe having our own explicit flag like `-on-dataflow-worker` is safer.

lock14:
@sethvargo @yolocs After reading the docs more thoroughly I discovered that it is required to call `beam.Init()` prior to doing any flag validation. The fix has been updated accordingly.

sethvargo (approved):

PR #171, opened 2023-11-02
Switch to use beam sdk logger

lock14:
* Log severity is not correctly categorized in Dataflow when using `pkg/logging`. Switching to use `beam/log` causes logs to show up correclty. See https://github.com/abcxyz/pkg/issues/236 for additional context.

### DIFF
--- a/pipeline/review/main.go
+++ b/pipeline/review/main.go
@@ -26,9 +26,9 @@ import (
 	"syscall"
 
 	"github.com/abcxyz/github-metrics-aggregator/pkg/review"
-	"github.com/abcxyz/pkg/logging"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/log"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam/x/beamx"
 )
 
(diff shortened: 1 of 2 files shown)


sethvargo:
Hey @lock14 - can you try using the text logger instead of the JSON logger?

```go
logger := logging.New(os.Stderr, slog.LevelDebug, logging.FormatText, true)
```

That _should_ output a similar-ish format. I'm perplexed about how it's parsing those text strings; must be a custom fluentd parser.

yolocs:
Beam seems to log to a beam specific logging server. It would be quite unfortunately if we need to diverge on the logging story

lock14:
I'm not sure if Dataflow changed something on there end, but I can't even get logs from `pkg/logging` logger to show up any more, only log messages from the `beam/log` are appearing
Code for logging test
image
Dataflow GCP log window:
image

lock14:
Examinging the logs a bit more, the `beam/log` logger appears to have a lot more fields in the `jsonPayload`
image
Whereas the `pkg/logging` logger only has `message` and `stream` fields in it's `jsonPayload`:
image

sethvargo:
This is super weird, but probably not worth spending more time debugging. Can you document why we're using a different logging inline in the code?

sethvargo (dismissed):

lock14:
> This is super weird, but probably not worth spending more time debugging. Can you document why we're using a different logging inline in the code?

done.

yolocs (approved):
It seems beam uses a different entirely different logging API. I can't really tell at which stage the logs get converted and put into cloud logging.

Could you update https://github.com/abcxyz/readability/wiki/Go-Best-Practices#logging to call out dataflow would need to use beam logging instead (and probably link to this PR for the findings you got). Thanks!

lock14:
This is just a hunch, but I suspect that the beam sdk may be using this function to instantiate a specially configured  logger when launched on a dataflow runner:
https://github.com/apache/beam/blob/master/sdks/go/pkg/beam/log/log.go#L62

I don't have any concrete evidence for this, but I thought I'd mention it in case someone else ever want to come back to this and dig some more.

lock14:
> It seems beam uses a different entirely different logging API. I can't really tell at which stage the logs get converted and put into cloud logging.
> 
> Could you update https://github.com/abcxyz/readability/wiki/Go-Best-Practices#logging to call out dataflow would need to use beam logging instead (and probably link to this PR for the findings you got). Thanks!

Updated the docs.

PR #172, opened 2023-11-02
WIP: improve pr view to not used received time to get latest informat…

pdewilde:
…ion about pr, but instead event time

### DIFF
--- a/terraform/modules/bigquery_metrics_views/data/bq_tvf/resources/pull_requests_by_date.sql
+++ b/terraform/modules/bigquery_metrics_views/data/bq_tvf/resources/pull_requests_by_date.sql
@@ -80,7 +80,7 @@ FROM
         INNER JOIN (
         SELECT
             id,
-            MAX_BY(delivery_id, received) delivery_id
+            MAX_BY(delivery_id, updated_at) delivery_id
         FROM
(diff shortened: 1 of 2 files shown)


terraform/modules/bigquery_metrics_views/data/bq_tvf/resources/pull_requests_by_date.sql
```diff
@@ -80,11 +80,11 @@ FROM
         INNER JOIN (
         SELECT
             id,
-            MAX(received) received
+            MAX_BY(delivery_id, updated_at) delivery_id
```
pdewilde, line 83: Ideally, it would be MAX_BY(deliveryid,  updated_at, received) as updated_at only has second granularity, so received would be able to break ties.

Also would be nice if a null updated_at was handled gracefully, but I don't think we do that upstream anyways, so its not really an issue.

lock14: `update_at` is listed as required in the GitHub docs. So I think it should never be null. https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request

Checking against our current data set, we don't have any pull request events with a null `updated_at`

PR #177, opened 2023-11-06
Create table for storing pull requests processed by future analyzer pipeline

stytchiz:


terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+  ])
+}
+
+resource "google_bigquery_table_iam_member" "invocation_comment_owners" {
+  for_each = toset(var.invocation_comment_table_iam.owners)
+
+  project = data.google_project.default.project_id
+
+  dataset_id = google_bigquery_dataset.default.dataset_id
+  table_id   = google_bigquery_table.leech_table.id
```
bradegler, line 792:
```suggestion
  table_id   = google_bigquery_table.invocation_comment_table.id
```

I think you wanted the new table, not the leech_table.

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+  role       = "roles/bigquery.dataOwner"
+  member     = each.value
+}
+
+resource "google_bigquery_table_iam_member" "leech_editors" {
```
bradegler, line 797: invocation_comment_editors

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+  for_each = toset(var.invocation_comment_table_iam.editors)
+
+  project = data.google_project.default.project_id
+
+  dataset_id = google_bigquery_dataset.default.dataset_id
+  table_id   = google_bigquery_table.leech_table.id
```
bradegler, line 803: invocation_comment_table.id again

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+  role       = "roles/bigquery.dataEditor"
+  member     = each.value
+}
+
+resource "google_bigquery_table_iam_member" "leech_viewers" {
```
bradegler, line 808: invocation_comment_viewers

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+  for_each = toset(var.invocation_comment_table_iam.viewers)
+
+  project = data.google_project.default.project_id
+
+  dataset_id = google_bigquery_dataset.default.dataset_id
+  table_id   = google_bigquery_table.leech_table.id
```
bradegler, line 814: invocation_comment_table.id again

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+  role       = "roles/bigquery.dataViewer"
+  member     = each.value
+}
```
bradegler, line 817: Please add a newline at the end of your files. Bonus points for configuring your editor to do it automatically.

bradegler (dismissed):

terraform/bigquery.tf
stytchiz: Yea whoops, I just noticed as well. For some reason, my git commits are failing probably because of gnubby auth issues.

```
➜  github-metrics-aggregator git:(stanchen/176) ✗ git commit -m "d'oh: fix some leftover string replacements"
Opening stream socket failed: null
error: Couldn't find key in agent?

fatal: failed to write commit object
```

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+      "description" : "URL of pull request."
+    },
+    {
+      "name" : "processed_at",
+      "type" : "TIMESTAMP",
+      "mode" : "REQUIRED",
+      "description" : "Timestamp of when the event was processed."
+    },
+    {
+      "name" : "status",
```
pdewilde, line 772: I think we should add a comment_id column in case we want to later go back and bulk-delete or bulk-modify comments we made using this pipeline. (For example if a bug is introduced that creates duplicate comments, or we migrate to a better solution and want to delete old comments).


According to the API docs its an int, though a string could also be used. https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+      "type" : "STRING",
```
pdewilde, line 773: Another field I considered but am less sold on is the message body, might make it easier if we change how things are done later down the line, but I'm like 50/50 whether that is a good idea.

pdewilde (commented):
We should probably discuss what we want to call this whole project. Invocation_comment seems like a good candidate, though I've been using pr_comment  or pr_log_comment in other places.

terraform/bigquery.tf
```diff
@@ -740,3 +740,78 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+      "description" : "ID of pull request."
+    },
+    {
+      "name" : "pull_request_html_url",
+      "type" : "STRING",
+      "mode" : "REQUIRED",
+      "description" : "URL of pull request."
+    },
+    {
+      "name" : "processed_at",
```
pdewilde, line 766: As well as processed_at, I was considering having either the received time of the merge event that we processed or the merge event, so we can know when to start processing from.

The only problem here will be dealing with failed runs. Best option I can think of so far is have some job that retries them all like once a week. (Alternatively we could have additional metadata about retry attemps and give up after some amount of tries).

terraform/bigquery.tf
```diff
@@ -740,3 +740,84 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+      "type" : "TIMESTAMP",
+      "mode" : "REQUIRED",
+      "description" : "Timestamp of when the event was processed."
```
pdewilde, line 769: I would specify processed by the invocation commenter

terraform/bigquery.tf
```diff
@@ -740,3 +740,84 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
...
+    },
+    {
+      "name" : "comment_id",
+      "type" : "INT64",
+      "mode" : "NULLABLE",
+      "description" : "ID of pull request comment."
+    },
+    {
```
pdewilde, line 777: What are your thoughts on having an "attempts" int column, so we can see how many times a failed PR was retried so we can eventually stop attempting.

stytchiz:
> We should probably discuss what we want to call this whole project. Invocation_comment seems like a good candidate, though I've been using pr_comment or pr_log_comment in other places.

@pdewilde I'm naming this table specifically for the use case of commenting on PRs with invocation artifacts (i.e. logs). We can rename the table to something more generic like `pr_comment` if it becomes more appropriate. 

I also named the pipeline the Analyzer pipeline. Please see the design doc for internal rationale and further discussion.

terraform/bigquery.tf
```diff
@@ -740,3 +740,90 @@ resource "google_bigquery_table_iam_member" "commit_review_status_viewers" {
   role       = "roles/bigquery.dataViewer"
   member     = each.value
 }
+
+# Invocation Comment Status Table / IAM
+
+resource "google_bigquery_table" "invocation_comment_table" {
```
bradegler, line 746: How does this table work? Are you writing a new row for each invocation attempt?

If so, then what is the purpose of the job_attempts column?

If your intention is to update this row with the latest job run then I would suggest rethinking that. BQ does not love updates.

pdewilde: job_attempts is for if you fail and need to retry. We don't want to retry forever, so after it is incremented a few times we will no longer attempt.

bradegler: So each run will lookup the previous run's value and increment it for this run? Why not use a select count by PR id?

pdewilde: That was my thought, but I guess select count by PR id would work too. It makes the selection query messier though.

bradegler: What criteria would you use to look up the last row? It would be similar, right? Select max(job_attempts) ...

Not sure one is better than the other but a count would eliminate the need to store the value.

pdewilde: So with the schema as it currently is, I can do something like

```sql
SELECT pull_request_id
FROM invocation_comment status
WHERE status = "SUCCESS" OR  job_attempts >= 3
```

to get the list of PRs I do not want to process, because they already succeeded or have failed too many times.

You could do the same with multiple rows, but the query would be more complex. You would need to group by pull_request_id, then have a having clause to limit to PRs with less than two rows, but I don't know how you could filter ones with success without a nested query of some sort.

pdewilde: https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax

Though I do see what you mean about updates being costly. `invocation_comment_status` should be a pretty small table (maybe like 10 MB) but maybe that isn't ideal.

pdewilde: I think I figured out a good query to do what I want
```sql
  HAVING COUNT(*) >= 3 OR COUNTIF(ics.status = "SUCCESS") > 0
  ```

pdewilde: Let's go ahead and remove this field, I think we can do without

pdewilde (approved):

PR #178, opened 2023-11-07
Add support for using a GitHub App for Auth

lock14:
* Added support for using a GitHub App to request an auth token instead of supplying one directly at startup
* Refactored code from `pkg/leech` involving fetching secrets from secret manager and requesting tokens using a GitHub App into their own packages so that they can be re-used.

### DIFF
--- a/pipeline/review/main.go
+++ b/pipeline/review/main.go
@@ -30,7 +30,6 @@ import (
 	"github.com/abcxyz/github-metrics-aggregator/pkg/auth"
 	"github.com/abcxyz/github-metrics-aggregator/pkg/review"
 	"github.com/abcxyz/github-metrics-aggregator/pkg/secrets"
-	"github.com/abcxyz/pkg/githubapp"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam/log"
@@ -99,11 +98,11 @@ func realMain(ctx context.Context) error {
 // using flag values. returns an error if any of the flags have malformed
 // data.
 func getConfigFromFlags(ctx context.Context) (*review.CommitApprovalPipelineConfig, error) {
-	githubAuther, err := getGitHubAuther(ctx)
+	githubTokenSupplier, err := getGitHubTokenSupplier(ctx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to construct token supplier: %w", err)
 	}
-	token, err := githubAuther.GitHubToken(ctx)
+	token, err := githubTokenSupplier.GitHubToken(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get GitHub token: %w", err)
 	}
@@ -147,14 +146,12 @@ func newQualifiedTableName(s string) (*bigqueryio.QualifiedTableName, error) {
 	return &bigqueryio.QualifiedTableName{Project: project, Dataset: dataset, Table: table}, nil
 }
 
-func getGitHubAuther(ctx context.Context) (GitHubAuther, error) {
+func getGitHubTokenSupplier(ctx context.Context) (auth.GitHubTokenSupplier, error) {
 	if githubToken != "" && githubAppID != "" {
 		return nil, fmt.Errorf("both a githubToken and githubAppID were supplied")
 	}
 	if githubToken != "" {
-		return &GitHubAuthToken{
-			token: githubToken,
-		}, nil
+		return auth.NewStaticGitHubTokenSupplier(githubToken), nil
 	}
 	if githubAppInstallationID == "" || githubAppPrivateKeyResourceName == "" {
 		return nil, fmt.Errorf("both githubAppInstallationID and githubAppPrivateKeyResourceName must be supplied when using a githubAppID")
@@ -163,39 +160,7 @@ func getGitHubAuther(ctx context.Context) (GitHubAuther, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to get private key: %w", err)
 	}
-	return &GitHubAuthApp{
-		githubAppID:             githubAppID,
-		githubAppInstallationID: githubAppInstallationID,
-		gitHubAppPrivateKey:     privateKey,
-	}, nil
-}
-
-type GitHubAuther interface {
-	GitHubToken(ctx context.Context) (string, error)
-}
-
-type GitHubAuthToken struct {
-	token string
-}
-
-func (g *GitHubAuthToken) GitHubToken(ctx context.Context) (string, error) {
-	return g.token, nil
-}
-
-type GitHubAuthApp struct {
-	githubAppID             string
-	githubAppInstallationID string
-	gitHubAppPrivateKey     *rsa.PrivateKey
-}
-
-func (g *GitHubAuthApp) GitHubToken(ctx context.Context) (string, error) {
-	githubAppConfig := githubapp.NewConfig(g.githubAppID, g.githubAppInstallationID, g.gitHubAppPrivateKey)
-	githubApp := githubapp.New(githubAppConfig)
-	token, err := auth.ReadAccessTokenForAllRepos(ctx, githubApp)
-	if err != nil {
-		return "", fmt.Errorf("failed to get GitHub access token: %w", err)
-	}
-	return token, nil
+	return auth.NewGitHubAppTokenSupplier(githubAppID, githubAppInstallationID, privateKey), nil
 }
 
 func getPrivateKey(ctx context.Context, secretResourceName string) (*rsa.PrivateKey, error) {
--- a/pkg/auth/githubapp.go
+++ b/pkg/auth/githubapp.go
@@ -16,6 +16,7 @@ package auth
 
 import (
 	"context"
+	"crypto/rsa"
 	"encoding/json"
 	"fmt"
 	"strings"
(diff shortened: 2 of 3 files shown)


pipeline/review/main.go
```diff
@@ -87,9 +96,20 @@ func realMain(ctx context.Context) error {
...
+		}
+		// empty githubToken, but non-empty GitHubApp flags, try requesting a token
+		// using the GitHubApp flags
+		token, err := requestToken(ctx)
+		if err != nil {
+			return nil, err
+		}
+		// successfully got the token, set githubToken to the value we got
+		githubToken = token
 	}
```
sethvargo, line 113: I'm not sure this is the right pattern. 

1. We should fail if a github token and appID are given, since it's not clear what behavior the user wants and it will lead to unexpected behaviors.

1. We should create an interface that vends a GitHub token, and have different implementations for "static" and "app". That will allow us to support other authentication in the future.


```go
type GitHubAuther interface {
  GitHubToken() (string, error)
}

type GitHubAuthToken struct {
  token string
}

func (g *GitHubAuthToken) GitHubToken() (string, error) {
  return g.token, nil
}

type GitHubAuthApp struct {
  githubAppID, githubAppInstallationID, githubAppSecret string
}

func (g *GitHubAuthApp) GitHubToken() (string, error) {
  // do dance
}
```

pipeline/review/main.go
```diff
@@ -110,3 +130,26 @@ func getConfigFromFlags() (*review.CommitApprovalPipelineConfig, error) {
...
+	}
+	privateKey, err := secrets.ParsePrivateKey(privateKeyString)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse private key: %w", err)
+	}
+	githubAppConfig := githubapp.NewConfig(githubAppID, githubAppInstallationID, privateKey)
+	githubApp := githubapp.New(githubAppConfig)
+	token, err := auth.ReadAccessTokenForAllRepos(ctx, githubApp)
+	if err != nil {
+		return "", fmt.Errorf("error getting GitHub access token: %w", err)
```
sethvargo, line 152:
```suggestion
		return "", fmt.Errorf("failed to get GitHub access token: %w", err)
```

pkg/auth/githubapp.go
```diff
@@ -0,0 +1,78 @@
...
+	// better to implement that in pkg so that GitHubTokenMinter can take
+	// advantage of it as well.
+	ghAppJWT, err := githubApp.AccessToken(ctx, tokenRequest)
+	if err != nil {
+		return "", fmt.Errorf("error creating GitHub access token: %w", err)
+	}
+	return parseJWT(ghAppJWT)
+}
+
+// ReadAccessTokenForAllRepos generate a new access token with read permissions
```
sethvargo, line 45: Isn't this duplicative with functionality in another repo already?

pkg/secrets/secrets.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to close secret manager client: %w", err)
+	}
+	return secret, nil
+}
+
+// ReadSecret reads a secret from Secret Manager using the give client and
+// validates that it was not corrupted during retrieval. The secretResourceName
+// should be in the format: 'projects/*/secrets/*/versions/*'.
+func ReadSecret(ctx context.Context, client *secretmanager.Client, secretResourceName string) (string, error) {
```
sethvargo, line 50: "Read" is not the correct name here, it's "Access". "Read" is the HTTP GET request that gets _metadata_, but not the value.

pkg/secrets/secrets.go
```diff
@@ -0,0 +1,78 @@
...
+	sm, err := secretmanager.NewClient(ctx)
+	if err != nil {
+		return "", fmt.Errorf("failed to create secret manager client: %w", err)
+	}
+	secret, err := ReadSecret(ctx, sm, secretResourceName)
+	if err != nil {
+		return "", fmt.Errorf("failed to retrieve secret: %w", err)
+	}
+	err = sm.Close()
+	if err != nil {
```
sethvargo, line 41:
```suggestion
	if err := sm.Close(); err != nil {
```

pkg/secrets/secrets.go
```diff
@@ -0,0 +1,78 @@
...
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+)
+
+// GetSecret reads a secret from Secret Manager and validates that it was not
+// corrupted during retrieval. The secretResourceName should be in the format:
+// 'projects/*/secrets/*/versions/*'.
+func GetSecret(ctx context.Context, secretResourceName string) (string, error) {
+	sm, err := secretmanager.NewClient(ctx)
```
sethvargo, line 32: It's _extremely_ expensive to create this client for each secret. The SecretManager client (like all clients) should be instantiated once and used repeatedly. It should also be closed during shutdown.

sethvargo (changes requested):

pkg/secrets/secrets.go
lock14: changed name to `AccessSecret`

pkg/auth/githubapp.go
lock14: I'm not aware of any other repo that has a similar function, was there one you had in mind?

pkg/secrets/secrets.go
lock14: There are two main scenarios I can think of:
1. Your program needs to access many secrets during the lifetime of its execution. In this case, I agree with you that a single client should be instantiated and re-used and only closed on shutdown. This is what the `AccessSecret` function in this file was meant to address where the secret manager client is passed in as a parameter.
2. Your program only needs to access one and only one secret during the lifetime of its execution. In this case, it makes sense to instantiate the client, fetch the secrete, then close the client since you no longer need it. This is use case the `GetSecret` function was meant to address.

Hopefully any users of these functions would be aware enough of their own programs needs to determine if `GetSecret` or `AccessSecret` should be used.

pipeline/review/main.go
lock14: I tried using this suggested interface structure. Let me know if this is what you had in mind or not.

pipeline/review/main.go
```diff
@@ -131,3 +146,66 @@ func newQualifiedTableName(s string) (*bigqueryio.QualifiedTableName, error) {
...
+	githubAppConfig := githubapp.NewConfig(g.githubAppID, g.githubAppInstallationID, g.gitHubAppPrivateKey)
+	githubApp := githubapp.New(githubAppConfig)
+	token, err := auth.ReadAccessTokenForAllRepos(ctx, githubApp)
+	if err != nil {
+		return "", fmt.Errorf("failed to get GitHub access token: %w", err)
+	}
+	return token, nil
+}
+
+func getPrivateKey(ctx context.Context, secreteResourceName string) (*rsa.PrivateKey, error) {
```
sethvargo, line 201: Typo

pkg/auth/githubapp.go
sethvargo: I could've sworn I saw a PR from @verbanicm that had this. Maybe it was the implementation for the client.

pkg/secrets/secrets.go
sethvargo: The `Get` versus `Access` semantic is extremely confusing because it diverges from the Secret Manager nomenclature. Get == Read == metadata. Access == secret material.

pkg/auth/githubapp.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	t, ok := tokenResp["token"]
+	if !ok {
+		return "", fmt.Errorf("malformed GitHub token response - missing token attribute")
+	}
+	token, ok := t.(string)
+	if !ok {
+		return "", fmt.Errorf("malformed GitHub token: wanted string got %T", token)
+	}
+	fmt.Printf("%s\n", tokenResp)
```
sethvargo, line 76: Debug?

pkg/auth/githubapp.go
```diff
@@ -0,0 +1,78 @@
...
+		return "", fmt.Errorf("error creating GitHub access token: %w", err)
+	}
+	return parseJWT(ghAppJWT)
+}
+
+// parseJWT parses the give JWT and returns the embedded token.
+func parseJWT(ghAppJWT string) (string, error) {
+	// The token response is a json doc with a lot of information about the
+	// token. All that is needed is the token itself.
+	var tokenResp map[string]any
```
sethvargo, line 64: It would be better to use a struct here and let the json package take care of the hard work for us:

```go
type ghTokenResponse struct {
  Token string `json:"token"`
}

var ght ghTokenResponse
if err := json.NewDecoder(strings.NewReader(ghAppJWT)).Decode(&ght); err != nil {
  return "", fmt.Errorf("failed to parse github token response: %w", err)
}

if ght.Token == "" {
  return "", fmt.Errorf("failed to parse github token response: no token in payload")
}
return ght.Token, nil
```

pkg/secrets/secrets.go
```diff
@@ -0,0 +1,77 @@
...
+	if checksum != *result.Payload.DataCrc32C {
+		return "", fmt.Errorf("failed to get privateKeyResourceName version for %q - data corrupted", secretResourceName)
+	}
+	return string(result.Payload.Data), nil
+}
+
+// ParsePrivateKey parses a string containing a PEM encoded private key into
+// a rsa.PrivateKey struct.
+func ParsePrivateKey(privateKeyContent string) (*rsa.PrivateKey, error) {
+	parsedKey, _, err := jwk.DecodePEM([]byte(privateKeyContent))
```
sethvargo, line 68: Why do we need the jwk library here? This is just an RSA key and we should be able to use https://pkg.go.dev/encoding/pem directly.

lock14: This is a fair call out. Since the function is only interested in parsing RSA private keys, we don't need the general nature of `jwk.DecodePEM`, I've rewritten the function to just use the `pem` package directly.

pkg/auth/githubapp.go
lock14: Done.

pkg/secrets/secrets.go
lock14: I was unaware of the `Get == Read == metadata` and `Access == secret material` distinction with regards to using secret manager. I've renamed the function to `AccessSecretFromSecretManager` to better indicate that we are actually getting the secret material. I've also extended the function doc to note that the `AccessSecret` function should be used if multiple secrets need to be fetched.

pkg/leech/ingest_logs.go
```diff
@@ -180,7 +179,7 @@ func (f *IngestLogsFn) ProcessElement(ctx context.Context, event EventRecord) Le
 // handleMessage is the main event processor. It generates a GitHub token, reads the workflow
 // log files if they exist and persists them to Cloud Storage.
 func (f *IngestLogsFn) handleMessage(ctx context.Context, repoName, ghLogsURL, gcsPath string) error {
-	token, err := f.repoAccessToken(ctx, repoName)
+	token, err := auth.ReadAccessTokenForRepos(ctx, f.ghApp, repoName)
```
pdewilde, line 182: Curious on thoughts for this, as I will be doing something similar. Does it make sense to just ask for read access to all repos and cache that in the local host?

In theory, Leech will be grabbing logs for an entire organization, so may different repositories may be accessed by a particular instance, so does it make sense to have O(repositories*runners) calls to grab granular tokens, or should we just grab a read all token once in each runner and use that for all calls.

@bradegler Might have opinions here?

pipeline/review/main.go
lock14: I've moved the interface and respective implementations to auth package.

pkg/leech/ingest_logs.go
lock14: This is essentially what I'm doing for the review pipeline. Since I need access to all repositories I simply use the GitHub App to fetch a token for all repos when pipeline starts up and just use that token throughout the pipeline.
https://github.com/abcxyz/github-metrics-aggregator/pull/178/files#diff-5b439c7a9f8f1115f2d8c05ba6b558ab32ffcde32aa55d3b18b746a64dd98bb7R105

sethvargo (approved):

pkg/leech/ingest_logs.go
pdewilde: So the token is passed along the pipeline then? For some reason I was thinking we would need to re-generate it at each step.

lock14: Yes, It gets stored in this config:
https://github.com/abcxyz/github-metrics-aggregator/blob/70753c04547ae797f337c82de9fdd24f059a5da9/pipeline/review/main.go#L122

Which is then used given to the beam DoFn structs:
https://github.com/abcxyz/github-metrics-aggregator/blob/70753c04547ae797f337c82de9fdd24f059a5da9/pkg/review/commit_review_status.go#L189

https://github.com/abcxyz/github-metrics-aggregator/blob/70753c04547ae797f337c82de9fdd24f059a5da9/pkg/review/commit_review_status.go#L204

PR #179, opened 2023-11-07
Use GoogleSQL syntax for table name flags

lock14:
* Switched from using `bigqueryio.NewQualifiedTableName` to parse table names from flags to new custom function `newQualifiedTableName`. `bigqueryio.NewQualifiedTableName` required BigQuery Legacy SQL syntax, whereas `newQualifiedTableName` uses GoggleSQL syntax.
* renamed `review.sqlFormat` to `review.formatGoogleSQL` to better reflect that the purpose of this function was to format the table name in GooogleSQL syntax.

### DIFF
--- a/pipeline/review/main_test.go
+++ b/pipeline/review/main_test.go
@@ -27,13 +27,13 @@ func TestNewQualifiedTableName(t *testing.T) {
 	cases := []struct {
 		name      string
 		tableName string
-		want      bigqueryio.QualifiedTableName
+		want      *bigqueryio.QualifiedTableName
 		wantErr   string
 	}{
 		{
 			name:      "parses_google_sql_format",
 			tableName: "project.dataset.table",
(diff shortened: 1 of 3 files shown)


pipeline/review/main_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/bigqueryio"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestNewQualifiedTableName(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name      string
+		tableName string
+		want      bigqueryio.QualifiedTableName
```
sethvargo, line 30: This should be a pointer

lock14: done.

sethvargo (approved):

PR #189, opened 2023-11-15
chore: add initial infra template

gjonathanhong:


abc.templates/infra/spec.yaml
```diff
@@ -0,0 +1,74 @@
...
+    params:
+      paths: ['contents']
+      as: ['{{.custom_name}}/infra']
+
+  - desc: 'Replace variables'
+    action: 'string_replace'
+    params:
+      paths: ['{{.custom_name}}/infra']
+      replacements:
+        - to_replace: 'REPLACE_CUSTOM_NAME'
```
verbanicm, line 55: Should we ensure lower hyphen case here?

verbanicm (approved):

PR #198, opened 2023-11-28
BigQuery identifier sanitization

pdewilde:
Google's golang BigQuery library does not support prepared statements, and the parameters are not able to be used for identifiers (parameters aren't even exposed for the beam runner).

This adds a validator for BigQuery identifiers to hopefully make string interpolation safer from injection risk. The risk model is not too dangerous for GMA pipelines run by a trusted admin, though if the existing pattern were copied to another project, it may have a different risk model, so it seems best to ensure safety here.

This PR doesn't fix the retry or webhook processors. Created #199 for it


pipeline/leech/main.go
```diff
@@ -138,6 +157,44 @@ func realMain(ctx context.Context) error {
 	return nil
 }
 
+// Validates inputs which are used in BigQuery string interpolation to help avoid SQL-Injection.
+func validateInputs(eventsProjectID *string, eventsTable *string, leechProjectID *string, leechTable *string) error {
+	var merr error
+	if err := bigquery.ValidateGCPProjectID(*eventsProjectID); err != nil {
+		merr = errors.Join(merr, fmt.Errorf("invalid events-project-id arg: %w", err))
+	}
+	eventsTableDataset, eventsTableName, err := splitDatasetTableName(*eventsTable)
```
drevell, line 166: This could perhaps be simplified by factoring out a function. The presence of "else" and the presence of copy-pasted blocks of code with variable names changed are both hints that factoring out a function could be beneficial. I'm thinking something like:

```
func splitAndValidate(dottedTableName string) (dataset string, table string, _ error) {
  dataset, table, err := splitDatasetTableName(*eventsTable)
  if err != nil {
    return err
  }
  return errors.Join(
    bigquery.ValidateDatasetID(eventsTableDataset),
    bigquery.ValidateTableName(eventsTableName))
}
```

And you might even be able to inline `splitDatasetTableName` into this function for more simplicity.

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"regexp"
+	"unicode/utf8"
+)
+
+// Start with lowercase, middle is lowercase, number or hyphen, cannot end in
+// hyphen. 6-30 characters in length (start, 4-28 middle, end).
+const projectIDRegex = `^[a-z][a-z0-9\-]{4,28}[a-z0-9]$`
+
+var projectIDMatcher = regexp.MustCompile(projectIDRegex)
```
drevell, line 28: Optional: it might be cleaner to inline projectIDRegex into the MustCompile() call.

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,99 @@
...
+// Unclear if these rules are valid for all external tables.
+// Does not check for restricted strings such as google, null, etc.
+func ValidateTableName(tableName string) error {
+	// TODO: is this really necessary, seems like regexp may do this for free.
+	if !utf8.Valid([]byte(tableName)) {
+		return fmt.Errorf("invalid table name: not UTF-8")
+	}
+	// Checking to ensure max 1024 UTF-8 bytes, as that is the limit actually
+	// in place.
+	if len(tableName) > 1024 {
```
drevell, line 90: Optional: `const maxTableUTF8Bytes = 1024`

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,319 @@
...
+// limitations under the License.
+
+package bigquery
+
+import (
+	"strings"
+	"testing"
+)
+
+func genStringLen(num int, char rune) string {
```
drevell, line 22: Perhaps https://pkg.go.dev/strings#Repeat?

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	var sb strings.Builder
+	sb.Grow(num)
+	for i := 0; i < num; i++ {
+		sb.WriteRune(char)
+	}
+	return sb.String()
+}
+
+func Test_ValidateGCPProjectID(t *testing.T) {
```
drevell, line 34: Conventionally there's not an underscore in a test name except when testing a method. https://github.com/abcxyz/readability/wiki/Go-Best-Practices#testing

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,319 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if err := ValidateGCPProjectID(tc.input); (err != nil) != tc.wantErr {
+				t.Errorf("ValidateGCPProjectID() error = %v, wantErr %v", err, tc.wantErr)
+			}
+		})
+	}
+}
+
+func Test_ValidateDatasetID(t *testing.T) {
```
drevell, line 153: Same here with the underscore, and elsewhere

drevell (changes requested):

pkg/bigquery/sanitize_test.go
pdewilde: I always forget all the parts I need to change from what my IDE generates, thanks!

pkg/bigquery/sanitize.go
pdewilde: I think I'll leave it as-is. The regex is already a long and hard to read line.

lock14 (dismissed):

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,99 @@
...
+const datasetIDRegex = `^[a-zA-Z0-9_]{1,512}[a-zA-Z0-9_]{0,512}$`
+
+var datasetIDMatcher = regexp.MustCompile(datasetIDRegex)
+
+// Unicode characters in category L (letter), M (mark), N (number),
+// Pc (connector, including underscore), Pd (dash), Zs (space). Max 1024
+// characters. 1024 is actually UTF-8 bytes from experimentation.
+// UTF-8 length check will be done separately.
+// regexp only allows 1000 repetitions, so had to manually repeat.
+const tableNameRegex = `^[\p{L}\p{M}\p{N}\p{Pc}\p{Pd}\p{Zs}]{1,512}[\p{L}\p{M}\p{N}\p{Pc}\p{Pd}\p{Zs}]{0,512}$`
```
drevell, line 41: Since the length is already enforced elsewhere, it might be unnecessary and potentially confusing to also enforce it here. Especially with the hack to work around the limitation of 1000 repetitions.

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,99 @@
...
+// Unclear if these rules are valid for all external tables.
+// Does not check for restricted strings such as google, null, etc.
+func ValidateTableName(tableName string) error {
+	// TODO: is this really necessary, seems like regexp may do this for free.
+	if !utf8.Valid([]byte(tableName)) {
+		return fmt.Errorf("invalid table name: not UTF-8")
+	}
+	// Checking to ensure max UTF-8 bytes, as that is the limit actually
+	// in place.
+	if len(tableName) > tableNameMaxUTF8Bytes {
```
drevell, line 90: If we also enforced len > 0 here, we could avoid dealing with length in the regex altogether.

drevell (changes requested):

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,99 @@
...
+// Returns nil if provided string is a valid table name, else returns an
+// error. Based on rules defined in
+// [https://cloud.google.com/bigquery/docs/tables#table_naming].
+// Identifiers cannot be replaced via parameters in BigQuery, so interpolation
+// is a necessry evil.
+// Hardcoding table names may be preferred to avoid any injection risk.
+// Unclear if these rules are valid for all external tables.
+// Does not check for restricted strings such as google, null, etc.
+func ValidateTableName(tableName string) error {
+	// TODO: is this really necessary, seems like regexp may do this for free.
```
stytchiz, line 84: TODOs should be imperative (e.g. TODO: `Do this when blah blah`). See internal resources for TODO style. Otherwise, just remove this altogether if there's not really any need for it.

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,307 @@
...
+			name:    "dot_fails",
+			input:   "github.metrics",
+			wantErr: true,
+		},
+		{
+			name:    "unicode_fails",
+			input:   "様様様様様様",
+			wantErr: true,
+		},
+		{
```
stytchiz, line 109: I think this is a duplicate of "empty_string_fails"

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,307 @@
...
+			input:   "foobar\nbarfoo",
+			wantErr: true,
+		},
+	}
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if err := ValidateGCPProjectID(tc.input); (err != nil) != tc.wantErr {
+				t.Errorf("ValidateGCPProjectID() error = %v, wantErr %v", err, tc.wantErr)
```
stytchiz, line 135: Include arg

`t.Errorf("ValidateGCPProjectID(%q) error = %v, wantErr %v", tc.input, err, tc.wantErr)`

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,307 @@
...
+			input:   "foobar\nbarfoo",
+			wantErr: true,
+		},
+	}
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if err := ValidateDatasetID(tc.input); (err != nil) != tc.wantErr {
+				t.Errorf("ValidateDatasetID() error = %v, wantErr %v", err, tc.wantErr)
```
stytchiz, line 229: Include arg here too

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,307 @@
...
+			input:   "foobar\nbarfoo",
+			wantErr: true,
+		},
+	}
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if err := ValidateTableName(tc.input); (err != nil) != tc.wantErr {
+				t.Errorf("ValidateTableName() error = %v, wantErr %v", err, tc.wantErr)
```
stytchiz, line 303: Include arg here too

pipeline/leech/main.go
```diff
@@ -139,6 +149,38 @@ func realMain(ctx context.Context) error {
...
+	if len(parts) != 2 {
+		return fmt.Errorf("expected 2 parts separated by a . but got %v", len(parts))
+	}
+	return errors.Join(
+		bigquery.ValidateDatasetID(parts[0]),
+		bigquery.ValidateTableName(parts[1]))
+}
+
+// Validates inputs which are used in BigQuery string interpolation to help avoid SQL-Injection.
+func validateInputs(eventsProjectID, eventsTable, leechProjectID, leechTable *string) error {
```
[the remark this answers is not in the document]
pdewilde: Copy paste error, will fix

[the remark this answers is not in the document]
pdewilde: You could, though you would lose the informative error message telling you which input was wrong, rather than the value of that input.

pkg/bigquery/sanitize_test.go
pdewilde: good catch

pkg/bigquery/sanitize.go
pdewilde: That's a good point. I'll change it.

pkg/bigquery/sanitize.go
pdewilde: I went ahead and added the check here, though the lower bound will also be checked because I did + at the end of the regex pattern.

pkg/bigquery/sanitize.go
pdewilde: Thanks, removed todo but left the check as its cheap

lock14 (approved):

drevell (dismissed):

pipeline/leech/main.go
```diff
@@ -139,6 +149,38 @@ func realMain(ctx context.Context) error {
 	return nil
 }
 
+func splitAndValidate(dottedTableName string) error {
+	parts := strings.Split(dottedTableName, ".")
```
sethvargo, line 153:
```suggestion
	parts := strings.SplitN(dottedTableName, ".", 2)
```

pipeline/leech/main.go
```diff
@@ -139,6 +149,38 @@ func realMain(ctx context.Context) error {
 	return nil
 }
 
+func splitAndValidate(dottedTableName string) error {
```
sethvargo, line 152: This is a fairly generic name for a function - split and validate what? At minimum, add godoc.

pipeline/review/main.go
```diff
@@ -165,5 +166,19 @@ func newQualifiedTableName(s string) (*bigqueryio.QualifiedTableName, error) {
 	if strings.TrimSpace(project) == "" || strings.TrimSpace(dataset) == "" || strings.TrimSpace(table) == "" {
 		return nil, fmt.Errorf("table name has empty components: %s", s)
 	}
+	var merr error
```
sethvargo, line 169: There are some existing validations above - can they be combined here?

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,98 @@
...
+// Package bigquery implements helpers for using BigQuery.
+package bigquery
+
+import (
+	"fmt"
+	"regexp"
+	"unicode/utf8"
+)
+
+const tableNameMaxUTF8Bytes = 1024
```
sethvargo, line 24: - godoc
- where does this come from? please link to docs or justification?

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,98 @@
...
+var datasetIDMatcher = regexp.MustCompile(datasetIDRegex)
+
+// Unicode characters in category L (letter), M (mark), N (number),
+// Pc (connector, including underscore), Pd (dash), Zs (space). Max 1024
+// characters. 1024 is actually UTF-8 bytes from experimentation.
+// UTF-8 length check will be done separately.
+// regexp only allows 1000 repetitions, so had to manually repeat.
+const tableNameRegex = `^[\p{L}\p{M}\p{N}\p{Pc}\p{Pd}\p{Zs}]+$`
+
+var tableNameMatcher = regexp.MustCompile(tableNameRegex)
```
sethvargo, line 43: These all seem to be focused on validation instead of extraction? I'm generally -1 on overzealous client-side validation, because it puts us in a precarious position when the server changes constraints. In general, we should let the server fail and give the user the appropriate response.

As written, if GCP allowed 30-character project IDs, our tool would have to change. I think that's a poor user experience. Same for the rest of these things. Client-side validation is rarely helpful.

I'm also not seeing how these regexes are protecting against injection attacks.

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"regexp"
+	"unicode/utf8"
+)
+
+const tableNameMaxUTF8Bytes = 1024
+
+// Start with lowercase, middle is lowercase, number or hyphen, cannot end in
+// hyphen. 6-30 characters in length (start, 4-28 middle, end).
+var projectIDMatcher = regexp.MustCompile(`^[a-z][a-z0-9\-]{4,28}[a-z0-9]$`)
```
sethvargo, line 28: If we keep these regexes, switch to `\A` and `\z`

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,98 @@
...
+// is a necessry evil.
+// In most cases, setting the default dataset in the client's Query object
+// is preferable as it avoids the injection risk entirely.
+// Does not check for restricted strings such as google, null, etc.
+func ValidateDatasetID(datasetID string) error {
+	if !datasetIDMatcher.MatchString(datasetID) {
+		return fmt.Errorf("invalid dataset id")
+	}
+	return nil
+}
```
sethvargo, line 73: These feel like strange helper functions - why not just make the regex a public variable?

pkg/bigquery/sanitize.go
```diff
@@ -0,0 +1,98 @@
...
+// Does not check for restricted strings such as google, null, etc.
+func ValidateTableName(tableName string) error {
+	if !utf8.Valid([]byte(tableName)) {
+		return fmt.Errorf("invalid table name: not UTF-8")
+	}
+	// Checking to ensure max UTF-8 bytes, as that is the limit actually
+	// in place.
+	if len(tableName) > tableNameMaxUTF8Bytes || len(tableName) < 1 {
+		return fmt.Errorf("invalid table name: too few/many bytes: got %v expected [1, %v]", len(tableName), tableNameMaxUTF8Bytes)
+	}
```
sethvargo, line 91:
```suggestion
	if l := len(tableName); l > tableNameMaxUTF8Bytes || ; < 1 {
		return fmt.Errorf("invalid table name: too few/many bytes: got %d expected [1, %d]", l, tableNameMaxUTF8Bytes)
	}
```

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,302 @@
...
+package bigquery
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestValidateGCPProjectID(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
```
sethvargo, line 24:
```suggestion
func TestValidateGCPProjectID(t *testing.T) {
	t.Parallel()
	
	cases := []struct {
```

pkg/bigquery/sanitize_test.go
```diff
@@ -0,0 +1,302 @@
...
+		},
+	}
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if err := ValidateGCPProjectID(tc.input); (err != nil) != tc.wantErr {
+				t.Errorf("ValidateGCPProjectID(%q) error = %v, wantErr %v", tc.input, err, tc.wantErr)
+			}
+		})
```
sethvargo, line 132:
```suggestion
	}
	
	for _, tc := range cases {
		tc := tc
		
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			
			if err := ValidateGCPProjectID(tc.input); (err != nil) != tc.wantErr {
				t.Errorf("ValidateGCPProjectID(%q) error = %v, wantErr %v", tc.input, err, tc.wantErr)
			}
		})
```

(throughout all tests)

sethvargo (changes requested):

pipeline/leech/main.go
pdewilde: I'm not sure I like that, as it shadows any invalid input.

For example datasetID.tablename.oops would be split into datasetID, tablename.oops with no error

pipeline/leech/main.go
pdewilde: Changed to `splitDatasetIDTableName`

pipeline/leech/main.go
```diff
@@ -139,6 +149,38 @@ func realMain(ctx context.Context) error {
...
+		merr = errors.Join(merr, fmt.Errorf("invalid leech-project-id arg: %w", err))
+	}
+	if err := splitAndValidate(leechTable); err != nil {
+		merr = errors.Join(merr, fmt.Errorf("invalid leech-table input: %w", err))
+	}
+
+	if merr != nil {
+		return fmt.Errorf("invalid inputs: %w", merr)
+	}
+	return nil
```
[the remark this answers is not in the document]
pdewilde: I guess I wanted it to be clear to the reader that it isn't an error case. Is there a reason to return `merr`?

pipeline/review/main.go
pdewilde: Hmm, I think I'll do a bit of a rewrite of this.

It doesn't handle invalid inputs with more than two `.` 

I'll make some changes and see how it looks

pkg/bigquery/sanitize.go
pdewilde: My assumption is that valid table names cannot produce an injection attack as we can be sure that there are no non-identifier characters. If that was possible, there would be some serious problems with table naming rules.

Another option would be to look for cases of "bad" strings such as `";` but disallowing all possible bad strings is a harder problem than only allowing known good strings.

pkg/bigquery/sanitize.go
pdewilde: I think I had two reasons:

1. The defensive java developer in me doesn't like exposing public variables, as it allows for external mutation. I get that's not really the go ethos though.

2. I wanted to document the rules I was enforcing, though I think I could just add to public var and it would be exposed by GODOC.

pkg/bigquery/sanitize.go
pdewilde: Though maybe just checking for no "`" chars is sufficient. I wasn't sure if there was some clever unicode trick or something that could get around it.

pipeline/leech/main.go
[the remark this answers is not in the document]
sethvargo: Sorry, I mean cut all lines 178-181 and `return merr`. This is currently the equivalent of:

```go
if false {
  return false
}
return true
```

pkg/bigquery/sanitize.go
sethvargo: I would probably make a reasonably positioned regular expression here, match against the table name, log a warning if it doesn't match, and then continue with processing.

pkg/bigquery/sanitize.go
sethvargo: Do you intend to use them in other packages? The private variable is still mutable within the package.

pipeline/leech/main.go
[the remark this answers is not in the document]
pdewilde: If I did that, I wouldn't be wrapping the errors in the merr, is that not the best practice?

pkg/bigquery/sanitize.go
pdewilde: I wanted to provide the option, in case someone wanted to reuse the logic, to make Public accessors to the individual validations (dataset, project, table ect)

pipeline/leech/main.go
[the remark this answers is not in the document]
sethvargo: Unless you're adding additionally helpful context, the wrapping probably isn't helpful.

pipeline/leech/main.go
sethvargo: That's fair.

pkg/bigquery/sanitize.go
sethvargo: It would probably make more sense to have that in pkg then. If we want something re-usable, let's build it that way. But I don't think we should make something public because someone _might_ use it: https://www.hyrumslaw.com/

pipeline/review/main.go
pdewilde: Committed a change, PTAL

PR #203, opened 2023-11-30
add user_project_override to infra template

gjonathanhong:


abc.templates/infra/contents/terraform.tf
```diff
@@ -31,4 +31,6 @@ terraform {
   }
 }
 
-provider "google" {}
+provider "google" {
+  user_project_override = true
```
verbanicm, line 35: Using this always makes me worried about downstream effects, but I get the ease of setup with it.

verbanicm (approved):

PR #204, opened 2023-12-04
add required IAM for actuating GMA terraform to infra template

gjonathanhong:


abc.templates/infra/contents/main.tf
```diff
@@ -41,3 +41,11 @@ module "REPLACE_MODULE_NAME" {
   leech_bucket_name     = local.leech_bucket_name
   leech_bucket_location = local.leech_bucket_location
 }
+
+module "REPLACE_MODULE_NAME_guardian_iam" {
+  source = "git::https://github.com/abcxyz/github-metrics-aggregator.git//terraform?ref=v0.0.7"
```
gjonathanhong, line 46: will bump version in follow up

verbanicm: Rather than having this be a module, should we just create an `iam.tf` file, then you can depend on the `google_project_iam_member.github_metrics_guardian_iam` instead. Since that resource depends on the custom roles, it should "just work".

Also, do we want to use the word Guardian here for resource names? There could be a possibility that people arent using Guardian which could be confusing.

abc.templates/infra/spec.yaml
```diff
@@ -32,6 +32,9 @@ inputs:
     desc: 'The CI service account email for deploying GitHub Metrics Aggregator'
     rules:
       - rule: 'gcp_matches_service_account(automation_service_account_email)'
+  - name: 'terraform_actuator_members'
```
gjonathanhong, line 35: dislike the name, but can't think of better

abc.templates/infra/contents/iam.tf
```diff
@@ -0,0 +1,75 @@
...
+    "roles/compute.networkAdmin",      # for creating the load balancer
+    "roles/compute.securityAdmin",     # to manage SSL certificates
+    "roles/iam.securityAdmin",         # set IAM policy on any resource within the project
+    "roles/iam.serviceAccountCreator", # allow for creation of service accounts
+    "roles/iam.serviceAccountUser",    # for deployment of initial cloud run image
+    "roles/iam.workloadIdentityPoolAdmin",
+    "roles/pubsub.editor",                  # create topics and subscriptions
+    "roles/run.admin",                      # create and manage cloud run services
+    "roles/serviceusage.serviceUsageAdmin", # enabled services on the project
+    google_project_iam_custom_role.cloudscheduler_job_creator.id,
```
verbanicm, line 30: Actually, I think this might not work because these resources are dynamic and terraform won't allow that in a for each IIRC.

You could hard code the role id instead and add a depends on.

abc.templates/infra/spec.yaml
verbanicm: Fine for now, do we need a list here tho? Ideally this is a single service account used by terraform. I'm not sure I have an opinion just posing the question.

abc.templates/infra/contents/iam.tf
gjonathanhong: you are right, nice catch!

abc.templates/infra/contents/iam.tf
```diff
@@ -0,0 +1,75 @@
...
+    "roles/serviceusage.serviceUsageAdmin", # enabled services on the project
+    "projects/${local.project_id}/roles/${local.cloudscheduler_job_creator}",
+    "projects/${local.project_id}/roles/${local.cloudstorage_bucket_creator}",
+    "projects/${local.project_id}/roles/${local.secretmanager_secret_creator}",
+  ])
+
+  project = local.project_id
+
+  role   = each.value
+  member = "REPLACE_TERRAFORM_ACTUATOR_MEMBER"
```
verbanicm, line 38: I think you need a depends on here also for each of the custom iam roles

verbanicm (approved):

PR #205, opened 2023-12-04
Update bigquery client interface to read and write for invocation comment statuses

stytchiz:


pkg/teeth/publish_logs.go
```diff
@@ -89,3 +101,52 @@ func SetUpPublisherSourceQuery(ctx context.Context, bqClient BigQueryClient) (*b
...
+// of the BigQuery client to configure batch inserts.
+//
+// Returns a list of errors where each error corresponds to the batch
+// insert in order.
+func SaveInvocationCommentStatus(ctx context.Context, bqClient BigQueryClient, statuses []*InvocationCommentStatusRecord) []error {
+	errors := make([]error, 0)
+	for i := 0; i <= len(statuses)/bqClient.InsertBatchSize(); i++ {
+		// Wait to prevent being rate-limited by BigQuery insert API.
+		if i > 0 {
+			time.Sleep(bqClient.InsertBatchWait())
```
sailorlqh, line 116: Could use `https://github.com/sethvargo/go-retry` instead of using `sleep` can be better?

Not a strong opinion, just pointing out a possible alternative.

stytchiz: Cool thanks! I'll definitely need this eventually.

So yes, a backoff implementation may work ... but (1) could be less efficient because the quota problem occurs when BigQuery batch inserts are done within a very short period of time - thundering herd (2) would require more error handling. 

Example: Constant backoff of 1s
`t+0s`: N calls happen, N-1 calls get some quota error or 429 Too Many Requests, gotta handle user error and have N-1 calls wait 1s ...
`t+1s`: N-1 calls happen, N-2 calls get user error, handle user error and have N-2 calls wait 1s ...
etc. ... not-to-mention the cases when 1 or calls get some internal error.

I see that https://github.com/sethvargo/go-retry also supports random jitters and max retries to help fix these issues. But if this is already sufficient and simple, then I will play the YAGNI card :)

pkg/teeth/publish_logs_test.go
```diff
@@ -124,3 +168,122 @@ ORDER BY
...
+		wantErrorCount int
+	}{
+		{
+			name:     "success",
+			bqClient: DefaultFakeBigQueryClient(),
+			statuses: []*InvocationCommentStatusRecord{
+				{
+					PullRequestID:  123456789,
+					PullRequestURL: "https://github.com/foo/bar/pull/1",
+					ProcessedAt:    time.Now(),
```
sethvargo, line 187: Always UTC your times

pkg/teeth/publish_logs_test.go
```diff
@@ -124,3 +168,122 @@ ORDER BY
...
+					CommentID:      bigquery.NullInt64{Int64: time.Now().Unix()},
+					Status:         "SUCCESS",
+					JobName:        "job-0",
+				},
+			},
+			wantErrorCount: 2,
+		},
+	}
+	for _, test := range tests {
```
sethvargo, line 271: Prefer `tc` over `test` for this variable name

pkg/teeth/publish_logs.go
stytchiz: That's a good point about context awareness. I added changes to handle cancellation + timeouts. PTAL.

pkg/teeth/publish_logs.go
```diff
@@ -35,11 +37,19 @@ import (
...
+	DefaultBQInsertBatchWait = 500 * time.Millisecond
+)
+
 // BigQueryClient defines the spec for calls to read from and write to
 // BigQuery tables.
 type BigQueryClient interface {
 	Config() *BQConfig
 	Query(string) *bigquery.Query
+	Insert([]*InvocationCommentStatusRecord) error
+	InsertBatchSize() int
```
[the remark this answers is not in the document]
stytchiz: Done.

pkg/teeth/publish_logs.go
```diff
@@ -89,3 +101,52 @@ func SetUpPublisherSourceQuery(ctx context.Context, bqClient BigQueryClient) (*b
...
 }
+
+// SaveInvocationCommentStatus inserts the statuses into the
+// InvocationCommentStatus table. It uses the batching parameters
+// of the BigQuery client to configure batch inserts.
+//
+// Returns a list of errors where each error corresponds to the batch
+// insert in order.
+func SaveInvocationCommentStatus(ctx context.Context, bqClient BigQueryClient, statuses []*InvocationCommentStatusRecord) []error {
+	errors := make([]error, 0)
```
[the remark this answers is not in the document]
stytchiz: My intention was to return a list of errors with each error corresponding to the batch insert operation. But returning a multi-error with `errors.Join` works well. Changed. Thanks!

pkg/teeth/publish_logs.go
```diff
@@ -89,3 +101,52 @@ func SetUpPublisherSourceQuery(ctx context.Context, bqClient BigQueryClient) (*b
...
+	for i := 0; i <= len(statuses)/bqClient.InsertBatchSize(); i++ {
```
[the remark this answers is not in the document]
stytchiz: Added variable.

pkg/teeth/publish_logs_test.go
stytchiz: Done

pkg/teeth/publish_logs_test.go
stytchiz: Done.

pkg/teeth/publish_logs.go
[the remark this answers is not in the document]
pdewilde: errors.Join() in this case may not be the best, as you will not be able to tell which values failed in a partial success, as errors.Join() ignores nil.

pkg/teeth/publish_logs.go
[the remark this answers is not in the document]
pdewilde: Might want to use a uint to avoid a class of bugs, though you still do have to check for zero, so not really more expensive than an int

pkg/teeth/publish_logs.go
```diff
@@ -99,29 +99,34 @@ func SetUpPublisherSourceQuery(ctx context.Context, bqClient BigQueryClient) (*b
...
+		if err := bqClient.Insert(ctx, statuses[start:end]); err != nil {
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				errs = append(errs, err)
+				break
+			}
+			errs = append(errs, fmt.Errorf("failed to insert batch #%d: %w", i, err))
 		}
 	}
-	return errors
+	return errors.Join(errs...)
```
pdewilde, line 129: If you continuing doing errors.Join() its probably more idiomatic to use the `merr` pattern

```go
var merr error
err := iMakeAnError()
if err != nil {
  merr = errors.Join(merr, fmt.Errorf("error from iMakeAnError(): %w", err))
}
...
return merr
```

pkg/teeth/publish_logs.go
yolocs: The sleep still doesn't handle cancellation / timeout properly. It would need something like:

```go
select {
case <-ctx.Done(): //context cancelled
case <-time.After(bqClient.InsertBatchWait()):
}
```

Using sethvargo/go-retry actually saves you the trouble to do this yourself.

pkg/teeth/publish_logs.go
```diff
@@ -87,5 +99,59 @@ func SetUpPublisherSourceQuery(ctx context.Context, bqClient BigQueryClient) (*b
...
+		// Wait to prevent being rate-limited by BigQuery insert API.
+		if i > 0 {
+			time.Sleep(bqClient.InsertBatchWait())
+		}
+		start := i * bqInsertBatchSize
+		end := min((i+1)*bqInsertBatchSize, len(statuses))
+		if err := bqClient.Insert(ctx, statuses[start:end]); err != nil {
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				errs = append(errs, err)
+				break
```
yolocs, line 124: On these errors, we want to abort the insertion right away. Why?

pkg/teeth/publish_logs_test.go
```diff
@@ -124,3 +169,141 @@ ORDER BY
...
+			if gotErrCount != tc.wantErrorCount {
+				t.Errorf("SaveInvocationCommentStatus returned %v, want %d errors", err, tc.wantErrorCount)
+			}
+			wantStatuses := make([]*InvocationCommentStatusRecord, 0)
+			if tc.wantErrorCount == 0 {
+				wantStatuses = tc.statuses
+			}
+			if diff := cmp.Diff(wantStatuses, tc.bqClient.processedStatuses); diff != "" {
+				t.Errorf("unexpected mismatch (-want +got):\n%s", diff)
+			}
```
yolocs, line 306: I think `cmp.Diff` should be able to handle comparing two empty slice so you don't need lines 300-303

yolocs (changes requested):

pkg/teeth/publish_logs_test.go
stytchiz: tc.statuses is not "gotStatuses". The insert behavior of the fake client is to store the input statuses in memory. But if the fake client is stubbed with an insertion error, it won't store the statuses and will return the stubbed error. I use `tc.wantErrorCount` to check for that since it's a simple proxy for expecting the error case.

pkg/teeth/publish_logs.go
stytchiz: On context cancellation or timeout, isn't it expected that the calling function should stop and clean up immediately ... therefore `break`? Let me know if I didn't get your point.

pkg/teeth/publish_logs.go
```diff
@@ -35,11 +37,19 @@ import (
...
+
 // BigQueryClient defines the spec for calls to read from and write to
 // BigQuery tables.
 type BigQueryClient interface {
 	Config() *BQConfig
-	Query(string) *bigquery.Query
+	Query(context.Context, string) *bigquery.Query
+	Insert(context.Context, []*InvocationCommentStatusRecord) error
+	InsertBatchSize() int64
+	InsertBatchWait() time.Duration
```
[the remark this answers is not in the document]
stytchiz: Fair point. Adding to `BQConfig`.

pkg/teeth/publish_logs.go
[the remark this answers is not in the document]
stytchiz: Yea I would think the error message will contain the failed values. The alternative would be to insert one-by-one but that's inefficient.

pkg/teeth/publish_logs.go
stytchiz: done.

pkg/teeth/publish_logs.go
stytchiz: Thanks. Added.

> Using sethvargo/go-retry actually saves you the trouble to do this yourself.
Sure but IIUC it'd still be different from what's implemented here. There'd be concurrent calls with some random backoff in the range of [backoff-jitter, backoff+jitter]. Here the calls are serialized. 

Honestly, I don't see a serious need for the batching and was doing this for defensive programming reasons whereby 1000s of PRs got merged into a single repo. 

Given how high BigQuery's insert quota/limits are, I'm gonna remove it all. If there's a discussion on needing it,  I can stack it in another PR.

pkg/teeth/publish_logs.go
stytchiz: Ended up removing.

stytchiz:
Got a lot of merge conflicts as a result of pulling from main. Seems like this branch was severely behind.

Opted to close this PR and start using git-town.

pkg/teeth/publish_logs.go
sethvargo: The calls would be serialized with go-retry too.

PR #207, opened 2023-12-06
refactor: move leech to separate module

gjonathanhong:
This change refactors the leech pipeline resources, used for GMA artifact management, into its own module. Instead of creating all leech-related resources by default, the leech pipeline is now an opt-in feature, by creating a configuration using the `leech` input variable.


verbanicm (approved):
I dont like the name leech, but thats a different discussion

PR #222, opened 2024-01-05
Add bigquery functionality for teeth job

stytchiz:
I'm splitting an [old PR](https://github.com/abcxyz/github-metrics-aggregator/pull/205) using git-town workflow. See https://github.com/abcxyz/github-metrics-aggregator/pull/205#issuecomment-1878071455.


pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,167 @@
...
+		LeechStatusTable             string
+	}
+	tablePrefix := fmt.Sprintf("%s.%s.", config.ProjectID, config.DatasetID)
+	fullTableNames := &Tables{
+		PullRequestEventsTable:       tablePrefix + config.PullRequestEventsTable,
+		InvocationCommentStatusTable: tablePrefix + config.InvocationCommentStatusTable,
+		EventsTable:                  tablePrefix + config.EventsTable,
+		LeechStatusTable:             tablePrefix + config.LeechStatusTable,
+	}
+	tmpl, err := template.New("publisher").Parse(PublisherSourceQuery)
```
pdewilde, line 147: (for other reviewers): BigQuery's go client doesn't support prepared statements, and while default dataset and project id can be set, and parameters can be used for things like a number replacement in a WHERE query, table name cannot be safely set.) Unfortunately, injection risk is a necessary evil, though with a rather insignificant threat model due to limited access to config, and limited permissions of the service account.

I do have a PR out to try and make it even safer: https://github.com/abcxyz/github-metrics-aggregator/pull/198 but have been busy recently and haven't responded to all the comments

pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// below copied from https://pkg.go.dev/cloud.google.com/go/bigquery#hdr-Querying
+	q := bq.client.Query(qStr)
+	it, err := q.Read(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read: %w", err)
+	}
+	results := make([]*PublisherSourceRecord, 0)
+	for {
+		r := &PublisherSourceRecord{}
```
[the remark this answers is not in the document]
sethvargo: No, but 

```suggestion
		var r PublisherSourceRecord
```

would be 😄

pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, fmt.Errorf("failed to populate query: %w", err)
+	}
+
+	// below copied from https://pkg.go.dev/cloud.google.com/go/bigquery#hdr-Querying
+	q := bq.client.Query(qStr)
+	it, err := q.Read(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read: %w", err)
+	}
+	results := make([]*PublisherSourceRecord, 0)
```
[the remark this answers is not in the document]
sethvargo:
```suggestion
	var results []*PublisherSourceRecord
```

would be idiomatic, unless you know the number of elements, in which case:

```suggestion
	results := make([]*PublisherSourceRecord, 0, NUM_OF_ELEMENTS)
```

would be idiomatic.

pkg/teeth/bigquery.go
sethvargo, line 147: Parsing this template on each invocation feels expensive. Can you parse it once at boot or client creation and re-use it?

pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,167 @@
...
+		LeechStatusTable:             tablePrefix + config.LeechStatusTable,
+	}
+	tmpl, err := template.New("publisher").Parse(PublisherSourceQuery)
+	if err != nil {
+		return "", fmt.Errorf("failed to set up sql template: %w", err)
+	}
+	var b bytes.Buffer
+	if err = tmpl.Execute(&b, fullTableNames); err != nil {
+		return "", fmt.Errorf("failed to execute sql template: %w", err)
+	}
```
sethvargo, line 154: You can cheat a bit here if you want:

```suggestion
	tablePrefix := fmt.Sprintf("%s.%s.", config.ProjectID, config.DatasetID)
	tmpl, err := template.New("publisher").Parse(PublisherSourceQuery)
	if err != nil {
		return "", fmt.Errorf("failed to set up sql template: %w", err)
	}
	var b bytes.Buffer
	if err = tmpl.Execute(&b, map[string]string{
		"PullRequestEventsTable":       tablePrefix + config.PullRequestEventsTable,
		"InvocationCommentStatusTable": tablePrefix + config.InvocationCommentStatusTable,
		"EventsTable":                  tablePrefix + config.EventsTable,
		"LeechStatusTable":             tablePrefix + config.LeechStatusTable,
	}); err != nil {
		return "", fmt.Errorf("failed to execute sql template: %w", err)
	}
```

pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,167 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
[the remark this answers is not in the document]
stytchiz: Done.

pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,167 @@
...
+	_ "embed"
+)
+
+// PublisherSourceQuery is the source query that teeth job pipeline
+// will use to publish results.
+//
+//go:embed sql/publisher_source.sql
+var PublisherSourceQuery string
+
+// TODO: Add query limit param.
```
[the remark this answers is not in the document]
stytchiz: Yes.

pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,167 @@
...
+// Close closes the BigQuery client.
+func (bq *BigQuery) Close() error {
+	if err := bq.client.Close(); err != nil {
+		return fmt.Errorf("failed to close client: %w", err)
+	}
+	return nil
+}
+
+// Config returns the BigQuery client config.
+func (bq *BigQuery) Config() *BQConfig {
```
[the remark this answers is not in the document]
stytchiz: Good catch! Ended up not needing it anymore. Removed.

pkg/teeth/bigquery.go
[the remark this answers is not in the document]
stytchiz: Done.

pkg/teeth/bigquery.go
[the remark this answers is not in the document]
stytchiz: Done

pkg/teeth/bigquery.go
stytchiz: Good catch! Moved to client creation.

pkg/teeth/bigquery.go
stytchiz: Nice! Thanks!

pkg/teeth/bigquery.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func populatePublisherSourceQuery(ctx context.Context, config *BQConfig) (string, error) {
+	tablePrefix := fmt.Sprintf("%s.%s.", config.ProjectID, config.DatasetID)
+	tmpl, err := template.New("publisher").Parse(PublisherSourceQuery)
+	if err != nil {
+		return "", fmt.Errorf("failed to set up sql template: %w", err)
+	}
+	var b bytes.Buffer
+	if err = tmpl.Execute(&b, map[string]string{
```
yolocs, line 104:
```suggestion
	if err := tmpl.Execute(&b, map[string]string{
```

pkg/teeth/publish_logs.go
```diff
@@ -18,74 +18,31 @@
...
-	JobName        string             `bigquery:"job_name"`
-}
-
-// SetUpPublisherSourceQuery converts the PublisherSourceQuery string into a
-// Query object that the BigQueryClient can then execute. It populates the
-// query parameters with the BigQuery config values.
-func SetUpPublisherSourceQuery(ctx context.Context, bqClient BigQueryClient) (*bigquery.Query, error) {
-	tmpl, err := template.New("publisher").Parse(PublisherSourceQuery)
+// GetLatestSourceRecords gets the latest publisher source records.
+func GetLatestSourceRecords(ctx context.Context, bqClient BigQueryClient) ([]*PublisherSourceRecord, error) {
```
yolocs, line 33: This func and the one below - I can't tell if they are necessary. They seem to be making a simple call the the bqClient.

stytchiz: It serves as an abstraction layer between the (yet-to-be-implemented) handler and bigquery client. Note that the bigquery client is passed in as an interface. This allows the handler tests to stub this client as a test double later on.

pkg/teeth/bigquery.go
stytchiz: Done

pkg/teeth/publish_logs.go
yolocs: Why can't you use the bigqueryclient interface as the abstraction layer? If you plan to add more logic here, then I can agree it's useful. Otherwise, I still don't understand the purpose.

yolocs (approved):

PR #224, opened 2024-01-09
Improve handling of null reviewDecision

lock14:
After more research into when a `null` reviewDecision can occur, I determined we can be a bit smarter about how we assign the review status. 

Whenever the reviewDecision is `null` it means the repository has not enabled the `Require pull request reviews before merging` branch protection feature. Soure: https://github.com/orgs/community/discussions/24375

In this scenario any 'approving' review is sufficient to determine if the code was approved.  Therefore, whenever the `reviewDecision` field is `null`, we check if any reviews have the `APPROVED` state, and set the `reviewDecision` to `APPROVED` if so, otherwise we set it to the default status of `UNKNOWN`.

### DIFF
--- a/pkg/review/commit_review_status.go
+++ b/pkg/review/commit_review_status.go
@@ -150,7 +150,12 @@ type PullRequest struct {
 	DatabaseID     githubv4.Int
 	Number         githubv4.Int
 	ReviewDecision githubv4.String
-	URL            githubv4.String
+	Reviews        struct {
+		Nodes []struct {
+			State githubv4.String
+		}
+	} `graphql:"reviews(first: 100)"`
+	URL githubv4.String
 }
 
 // BreakGlassIssueFetcher fetches break glass issues from a data source.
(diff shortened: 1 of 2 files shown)


pkg/review/commit_review_status.go
```diff
@@ -488,8 +493,22 @@ func GetPullRequestsTargetingDefaultBranch(ctx context.Context, client *githubv4
 		// Only select pull requests made against the default branch.
 		for _, pr := range query.Repository.Object.Commit.AssociatedPullRequest.Nodes {
 			if pr.BaseRefName == query.Repository.DefaultBranchRef.Name {
+				// if review decision is not set (e.g. it is blank or null), then the
+				// repository does not have the 'Require pull request reviews before merging'
+				// branch protection setting is enabled.
+				// https://github.com/orgs/community/discussions/24375
```
yolocs, line 499: I'm OK with this but I wonder if we should have a different status to reflect such approvals? Just so that we know there is a nuance in these approvals.

lock14: I like the idea of distinguishing this scenario from other 'APPROVED' and 'UNKNOWN' cases. I'm writing up a one pager so we can get consensus on exactly what the names and procedure should be.

sethvargo: Sounds good - I agree this is a weird edge case. Thanks for writing something up so we can all align @lock14

lock14: I am closing as using `reviewDecision` was deemed to un-reliable as it can be set to null post-merge of a PR.
We are instead switching to using the underlying review data to determine a commit's approval status.
See  #231 for this change.

PR #226, opened 2024-01-10
Remove predefined roles and upgrade template version to v0.0.15

gjonathanhong:


abc.templates/infra/contents/main.tf
```diff
@@ -4,12 +4,11 @@ locals {
 }
 
 module "REPLACE_MODULE_NAME" {
-  source = "git::https://github.com/abcxyz/github-metrics-aggregator.git//terraform?ref=v0.0.14"
+  source = "git::https://github.com/abcxyz/github-metrics-aggregator.git//terraform?ref=v0.0.15"
 
   project_id = local.project_id
 
   image                             = "gcr.io/cloudrun/placeholder@sha256:f1586972ac147796d60ee2c5a0d6cc78067fc862b6d715d6d2a96826455c3423"
-  prefix_name                       = "REPLACE_CUSTOM_NAME"
```
gjonathanhong, line 12: Removing because prefix has 17 character limit due to wif service account name limitation of 28 characters.
It's unlikely that there will be multiple GMA instances, so I think we can just accept the default prefix.

abc.templates/infra/contents/iam.tf
verbanicm, line 1: Should we just co-locate all iam permissions in a central place? Seems strange to have some in the infra repo and some here?

gjonathanhong: Moved all IAM related resources to manual step

verbanicm (approved):

PR #228, opened 2024-01-10
Add Github logic to teeth for creating the invocation comments

stytchiz:


pkg/teeth/github.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// GitHub is just a wrapper around githubclient.GitHub for functionality specific to teeth.
+// Use NewGitHub to initialize this.
+type GitHub struct {
+	*githubclient.GitHub
+
+	config *GHConfig
+	// can be overriden for testing
+	commentPRFunc func(ctx context.Context, owner, repo string, prNumber int, comment string) (*github.Response, error)
```
sethvargo, line 50: For fields specific to testing that are embedded on a structure, prefer prefixing with `test`:

```suggestion
	testCommentPRFunc func(ctx context.Context, owner, repo string, prNumber int, comment string) (*github.Response, error)
```

sethvargo, line 50: For fields specific to testing that are embedded on a structure, prefer prefixing with `test`:

```suggestion
	testCommentPRFunc func(ctx context.Context, owner, repo string, prNumber int, comment string) (*github.Response, error)
```

sethvargo, line 50: For fields specific to testing that are embedded on a structure, prefer prefixing with `test`:

```suggestion
	testCommentPRFunc func(ctx context.Context, owner, repo string, prNumber int, comment string) (*github.Response, error)
```

pkg/teeth/github.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func NewGitHub(ctx context.Context, config *GHConfig, appID, rsaPrivateKeyPEM string) (*GitHub, error) {
```
sethvargo, line 53: Comment

pkg/teeth/github.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, fmt.Errorf("failed to setup github client: %w", err)
+	}
+	return &GitHub{
+		GitHub:        client,
+		config:        config,
+		commentPRFunc: client.CommentPR,
+	}, nil
+}
+
+func convertRetryable(statusCode int) error {
```
sethvargo, line 65: Naming nit: `shouldRetry`? It feels weird that this returns an error instead of a boolean too.

pkg/teeth/github.go
```diff
@@ -0,0 +1,94 @@
...
+func (gh *GitHub) CreateInvocationCommentWithRetry(ctx context.Context, prNum int, invocationComment string) error {
+	backoff := retryFunc(retryMinWaitDuration)
+	backoff = retry.WithMaxRetries(retryMaxAttempts, backoff)
+	if err := retry.Do(ctx, backoff, func(ctx context.Context) error {
+		resp, err := gh.commentPRFunc(ctx, gh.config.Owner, gh.config.Repo, prNum, invocationComment)
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+		return convertRetryable(resp.StatusCode)
```
sethvargo, line 89: It's not specific to GitHub, but if you want to borrow from: https://github.com/abcxyz/pkg/blob/main/gcpmetadata/client.go#L192

pkg/teeth/github_test.go
```diff
@@ -0,0 +1,72 @@
...
+				}}, nil
+			},
+			wantErr: true,
+		},
+	}
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			client := &GitHub{
```
sethvargo, line 62:
```suggestion
	}
	
	for _, tc := range tests {
		tc := tc
		
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			
			client := &GitHub{
```

pkg/teeth/github.go
pdewilde: Feels like this should maybe live in `pkg/githubclient`

pkg/teeth/github.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/abcxyz/github-metrics-aggregator/pkg/githubclient"
+	"github.com/google/go-github/v56/github"
+	"github.com/sethvargo/go-retry"
+)
+
+const ()
+
+var (
+	// can be overriden for testing
+	retryMinWaitDuration        = 1 * time.Second
```
pdewilde, line 32: I recall GitHub having org-wide rate limits for comment API, so just be careful when metering requests that you take into account retries

pkg/teeth/github.go
```diff
@@ -0,0 +1,94 @@
...
+// to the PR by number.
+//
+// It leverages a default retry policy to get around GitHub's API rate limits.
+func (gh *GitHub) CreateInvocationCommentWithRetry(ctx context.Context, prNum int, invocationComment string) error {
+	backoff := retryFunc(retryMinWaitDuration)
+	backoff = retry.WithMaxRetries(retryMaxAttempts, backoff)
+	if err := retry.Do(ctx, backoff, func(ctx context.Context) error {
+		resp, err := gh.commentPRFunc(ctx, gh.config.Owner, gh.config.Repo, prNum, invocationComment)
+		if err != nil {
+			return err
```
pdewilde, line 86: Shouldn't this be retryable? I would imagine it could just be a temporary network blip or timeout

pkg/teeth/github.go
stytchiz: Hmm feels weird to make a call to `gh.testCommentPRFunc` in `CreateInvocationCommentWithRetry` below? Is this a pattern or an optional suggestion?

pkg/teeth/github.go
stytchiz: Done

pkg/teeth/github.go
stytchiz: > Naming nit: shouldRetry? It feels weird that this returns an error instead of a boolean too.
Done. Also agree that `shouldRetry(...) bool` is more customary .

> Feels like this should maybe live in pkg/githubclient
Good point. I thought about it but pkg/githubclient has some other logic for the GMA retry service as well and I'm not familiar with the calls it makes. I lean toward giving the caller responsibility on which statuses to retry based on the specific Github API and scoping the githubclient to just handle requests and responses.

pkg/teeth/github_test.go
stytchiz: Done

pkg/teeth/github.go
stytchiz: Yea these errors occur when the request does not reach GitHub. I looked into what these errors could be. It turns out that rate limit errors are returned here: https://github.com/google/go-github/blob/e9f52699f5e55a4791567bce668f56446e6343ef/github/github.go#L1108

Will fix the above, but for timeouts and network blips (I expect both results in context deadline exceeded error), that probably indicates an outage. I wouldn't think it's useful retrying until the next time the job runs.

pkg/teeth/github.go
stytchiz: Ack see my other comment about RateLimitErrors. Will resolve there.

pkg/teeth/github.go
stytchiz: This is great! Added changes. Thanks!

pkg/teeth/github.go
sethvargo: This is a pattern we follow for fields that are only for testing.

sethvargo (dismissed):

pkg/teeth/github.go
```diff
@@ -81,14 +73,66 @@ func (gh *GitHub) CreateInvocationCommentWithRetry(ctx context.Context, prNum in
...
 	}); err != nil {
 		return fmt.Errorf("failed to create invocation comment: %w", err)
 	}
+
 	return nil
 }
+
+func shouldRetry(statusCode int, err error) bool {
+	// Retry if rate-limited. See https://github.com/google/go-github#rate-limiting.
+	if GitHubErrRateLimit.Is(err) || GitHubErrSecondaryRateLimit.Is(err) {
```
pdewilde, line 120: I don't think this is what you want. It is testing for equality while you want to test for type.

See the implmentation: https://github.com/google/go-github/blob/eb3bed00ca77f6e718f034d07008e6f7cfd93b17/github/github.go#L1162 It is checking for RetryAfter field matching.

If you look at docs for the library, it reccomends handling using type assertions instead https://pkg.go.dev/github.com/google/go-github/v58/github#hdr-Rate_Limiting

pkg/teeth/github.go
pdewilde: Yeah, that makes sense for a batch job. I'm more used real-time usecases

PR #233, opened 2024-01-16
update abc.templates to use _git_tag and _git_sha for at-render-time versioning

gjonathanhong:


pdewilde (approved):

abc.templates/infra/spec.yaml
```diff
@@ -70,3 +70,25 @@ steps:
           with: '{{.terraform_state_bucket}}'
         - to_replace: 'REPLACE_BUCKET_PREFIX'
           with: '{{toLowerHyphenCase .custom_name}}/infra'
+        - to_replace: '# REPLACE_GITHUB_METRICS_VERSION_TAG'
+          with: |-
+            {{- if ._git_tag -}}
+              # {{._git_tag}}
+            {{- else -}}
+              {{._git_tag}}
+            {{- end }}
```
verbanicm, line 79: Is this right? I read this as, if we have a value for git tag do nothing, otherwise, use git tag...

gjonathanhong: Yes, so I'm doing a string match `# REPLACE_GITHUB_METRICS_VERSION_TAG` to replace the existing comment.

If `_git_tag` exists, then we rewrite a new comment with the tag value, else we just replace the existing comment with an empty string.

verbanicm: AHHH syntax highlighting got me lol. Looks good, thanks for clarifying!

verbanicm (approved):

PR #241, opened 2024-01-23
Output commit review status table id

lock14:
Output the commit review status table id so that we can refer to it in the infra-gcp terraform

### DIFF
--- a/terraform/outputs.tf
+++ b/terraform/outputs.tf
@@ -73,7 +73,7 @@ output "bigquery_unique_events_view_id" {
 
 output "bigquery_commit_review_status_table_id" {
   description = "BigQuery commit_review_status table resource."
-  value       = var.commit_review_status.enabled ? module.commit_review_status[0].commit_review_status_table_id : null
+  value       = try(module.commit_review_status[0].commit_review_status_table_id, null)
 }
 
 output "bigquery_event_views" {
--- a/terraform/outputs.tf
+++ b/terraform/outputs.tf
@@ -71,6 +71,11 @@ output "bigquery_unique_events_view_id" {
   value       = google_bigquery_table.unique_events_view.table_id
 }
 
+output "bigquery_commit_review_status_table_id" {
+  description = "BigQuery commit_review_status table resource."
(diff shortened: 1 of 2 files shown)


terraform/outputs.tf
```diff
@@ -71,6 +71,11 @@ output "bigquery_unique_events_view_id" {
   value       = google_bigquery_table.unique_events_view.table_id
 }
 
+output "bigquery_commit_review_status_table_id" {
+  description = "BigQuery commit_review_status table resource."
+  value       = var.commit_review_status.enabled ? module.commit_review_status.commit_review_status_table_id : null
```
gjonathanhong, line 76: You may need to add `commit_review_status_table_id` as an output to the commit_review_status module so that you can access it here

gjonathanhong, line 76: Because the module uses the `count` meta argument [here](https://github.com/abcxyz/github-metrics-aggregator/blob/main/terraform/bigquery.tf#L526), you will need to add an index value here.

```suggestion
  value       = var.commit_review_status.enabled ? module.commit_review_status[0].commit_review_status_table_id : null
```

**nit** Additionally, because the module depends on the enabled flag, I think you could even "simplify" this to
```suggestion
  value       = try(module.commit_review_status[0].commit_review_status_table_id, null)
```

verbanicm: @lock14 Did you address this comment? I dont see the output in the codebase...

https://github.com/search?q=repo%3Aabcxyz%2Fgithub-metrics-aggregator%20commit_review_status_table_id&type=code

gjonathanhong (dismissed):

terraform/outputs.tf
verbanicm: Just curious, if we are just outputting the input, why do this needle threading?

verbanicm: Can we use the property from the bigqery table resource instead?

lock14: @verbanicm that makes more sense. done.

gjonathanhong (approved):

PR #246, opened 2024-01-26
Add code audit dashboard terraform

gjonathanhong:
- Renames input var from `commit_review_stats` to `code_audit_dashboard`.
- Adds VPC terraform module (for dataflow dependency)
- Adds IAM for viewing through Looker Studio dashboard

### DIFF
--- a/abc.templates/infra/contents/main.tf
+++ b/abc.templates/infra/contents/main.tf
@@ -31,6 +31,12 @@ module "REPLACE_MODULE_NAME" {
     editors = []
     viewers = []
   }
+  code_audit_dashboard = {
+    enabled = false
+    table_iam = {
+      viewers = []
+    }
+  }
   github_metrics_dashboard = {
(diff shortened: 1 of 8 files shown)


abc.templates/infra/contents/main.tf
```diff
@@ -31,6 +31,12 @@ module "REPLACE_MODULE_NAME" {
     editors = []
     viewers = []
   }
+  code_audit_dashboard = {
+    enabled = false
+    table_iam = {
+      viewers = []
+    }
```
verbanicm, line 38: nit: why is this different from github_metrics_dashboard inputs? Should we try to unify on a common set of inputs here for dashboards? The might require some further discussion...

gjonathanhong: I'm in agreement, they should have similar configurations. But that might require more refactoring, we can discuss this offline

PR #248, opened 2024-03-06
Update deps

sethvargo:


.github/workflows/cleanup.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         uses: 'docker://us-docker.pkg.dev/gcr-cleaner/gcr-cleaner/gcr-cleaner-cli' # ratchet:exclude old image versions are removed
         with:
           args: >-
-            -repo="${{ env.GAR_REPO }}"
+            -repo="${{ env.GAR_REPO }}" -recursive="true" -grace="336h" -tag-filter-any="(?i)[0-9a-f]{40}(-amd64|-arm64)?"
```
verbanicm, line 90: This looks incorrect since these flags are still below this line, assuming a ratchet issue?

sethvargo: Yea - ratchet doesn't handle that well...

Also, we should switch to GAR cleanup policies instead of doing this :)

verbanicm: Next PR

sailorlqh: > Also, we should switch to GAR cleanup policies instead of doing this :)

Thanks for letting me know cleanup_policy is no long only in beta provider.

PR #279, opened 2024-06-24
feat: refactored review pipeline to remove dependency on Apache Beam

bradegler:


pkg/review/job.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/abcxyz/github-metrics-aggregator/pkg/bq"
+	"github.com/abcxyz/github-metrics-aggregator/pkg/version"
+	"github.com/abcxyz/pkg/githubauth"
+	"github.com/abcxyz/pkg/logging"
+	"github.com/abcxyz/pkg/workerpool"
+)
+
+// ExecuteJob runs the pipeline job to read GitHub
+// commits to check if they were properly reviewed.
+func ExecuteJob(ctx context.Context, cfg *Config) error {
```
sethvargo, line 31: What happens if this job fails halfway though?

bradegler: Status of each commit is stored at the end, if the job fails in the middle it will not write updated status and it will start over on next run. This is the same functionality as the original Beam implementation.

sethvargo: Okay cool. I just want to make sure we don't (more easily) end up in a half-baked state because of differences between Beam and Cloud Run Jobs.

drevell:
Removed myself as reviewer since Seth is on it. LMK if you'd still like a review.

sethvargo (approved):

PR #286, opened 2024-07-15
Add Cloud Scheduler Terraform

lock14:
Added Cloud Scheduler Terraform configuration for the artifacts and commit review status jobs.

### DIFF
--- a/terraform/modules/commit_review_status/cloud_run_job.tf
+++ b/terraform/modules/commit_review_status/cloud_run_job.tf
@@ -76,3 +76,38 @@ resource "google_project_iam_member" "default" {
   member = google_service_account.default.member
   role   = "roles/secretmanager.secretAccessor"
 }
+
+// Give the scheduler invoker permission to the cloud run instance
+resource "google_cloud_run_service_iam_member" "invoker" {
+  project = var.project_id
+
+  location = var.region
+  service  = google_cloud_run_v2_job.default.name
+  role     = "roles/run.invoker"
+  member   = google_service_account.default.member
+
+  depends_on = [
+    google_cloud_scheduler_job.scheduler
+  ]
+}
+
+resource "google_cloud_scheduler_job" "scheduler" {
+  project = var.project_id
+
+  name             = "artifacts-job"
+  region           = var.region
+  schedule         = var.scheduler_cron
+  time_zone        = var.scheduler_timezone
+  attempt_deadline = var.scheduler_deadline_duration
+  retry_config {
+    retry_count = var.scheduler_retry_limit
+  }
+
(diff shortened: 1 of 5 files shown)


sethvargo (dismissed):

terraform/modules/commit_review_status/cloud_run_job.tf
verbanicm, line 1: It doesn't look like we are associating these Cloud Run jobs with a service account? We create a service account and it is used as invoker and from cloud scheduler, but should we use a custom service account for the job instead of the default one?

lock14: Each cloud run job has its own service account already defined. 

Here is the one for the artifacts job:
https://github.com/abcxyz/github-metrics-aggregator/blob/2d310cf75f2625e7ad43eae56c90eb8981003f4f/terraform/modules/artifacts/cloud_run_job.tf#L67C1-L71C2

And the commit review status job:
https://github.com/abcxyz/github-metrics-aggregator/blob/2d310cf75f2625e7ad43eae56c90eb8981003f4f/terraform/modules/commit_review_status/cloud_run_job.tf#L67C1-L71C2

It seemed fine to use these same same service accounts as the ones to invoke their respective jobs via cloud scheduler. Do you think we need a separate service account just for the cloud scheduler part as well?

verbanicm: I dont see that service account associated to the cloud run job tho?

lock14: Ah, you are right. I didn't author the cloud run job terraform, so I thought they had already been associated with the service account. I've updated them to use the service account now.

verbanicm: ack, we should have the discussion on if we should have two separate SAs, but we can do that offline

terraform/modules/commit_review_status/cloud_run_job.tf
```diff
@@ -76,3 +77,38 @@ resource "google_project_iam_member" "default" {
...
+
+  depends_on = [
+    google_cloud_scheduler_job.scheduler
+  ]
+}
+
+resource "google_cloud_scheduler_job" "scheduler" {
+  project = var.project_id
+
+  name             = "artifacts-job"
```
verbanicm, line 98: I assume this is a copy paste error?

lock14: yes, I actually meant to use the `var.job_name` instead of hardcoding it. updated now.

verbanicm (approved):

PR #288, opened 2024-07-17
Fix job name formatting

lock14:
Must match regex `^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$`

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -375,13 +375,13 @@ variable "github_metrics_dashboard" {
 variable "artifacts_job_name" {
   description = "The name to give the artifacts Cloud Run job"
   type        = string
-  default     = "artifacts_job"
+  default     = "artifacts-job"
 }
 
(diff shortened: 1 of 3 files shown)


terraform/variables.tf
```diff
@@ -375,13 +375,13 @@ variable "github_metrics_dashboard" {
 variable "artifacts_job_name" {
   description = "The name to give the artifacts Cloud Run job"
   type        = string
-  default     = "artifacts_job"
+  default     = "artifacts-job"
```
verbanicm, line 378: Should we add a regex validation to prevent future errors here?

verbanicm (dismissed):

terraform/variables.tf
lock14: good idea. done.

sethvargo (approved):

verbanicm (approved):

PR #295, opened 2024-07-23
Fix terraform default values

lock14:
Values were accidently swapped for `leech` and `commit_review_status` jobs


sethvargo (dismissed):

verbanicm (approved):
if possible, we should remove all references to "leech" but thats a super nit

bradegler (approved):

PR #309, opened 2024-07-29
fix use of deprecated field

lock14:
Github has deprecated the `databaseId` field for pull requests in their GraphQL API. They have replaced it with a new field called `fullDataBaseId` that may contain values that exceed the size of a 32-bit integer. This field is serialized as a string.

see: https://docs.github.com/en/graphql/reference/objects#pullrequest for more details

### DIFF
--- a/pkg/review/commit_review_status_test.go
+++ b/pkg/review/commit_review_status_test.go
@@ -112,7 +112,7 @@ func TestGetPullRequests(t *testing.T) {
 			want: []*PullRequest{
 				{
 					BaseRefName:    "main",
-					FullDatabaseID: "1",
+					FullDatabaseID: "8294967296",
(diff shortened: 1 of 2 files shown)


pkg/review/commit_review_status_test.go
```diff
@@ -675,9 +675,9 @@ func TestGetPullRequests(t *testing.T) {
 			},
 			want: []*PullRequest{
 				{
-					BaseRefName: "main",
-					DatabaseID:  1,
-					Number:      23,
+					BaseRefName:    "main",
+					FullDatabaseID: "1",
```
yolocs, line 679: Nit, somewhere in a test case, maybe use a BigInt number.

yolocs (approved):

PR #313, opened 2024-08-07
Extract worker pool logic into a reusable function

lock14:


pkg/review/job.go
```diff
@@ -81,61 +75,78 @@ func ExecuteJob(ctx context.Context, cfg *Config) error {
...
+	// statuses that should not be further processed will be nil, so we should exclude them
+	commitReviewStatuses = removeNil(commitReviewStatuses)
+
+	// Step 3: Look up break glass issue if necessary and tag the review status with it if found.
+	fetcher := &BigQueryBreakGlassIssueFetcher{
+		client: bqClient,
+	}
+	taggedReviewStatuses, err := pooledTransform(ctx, commitReviewStatuses,
+		func(status *CommitReviewStatus) (*CommitReviewStatus, error) {
+			return processReviewStatus(ctx, fetcher, cfg, *status), nil
```
verbanicm, line 95: Is there a reason to have the processReviewStatus func deference the status var? Can we pass the pointer instead?

pkg/review/job.go
```diff
@@ -81,61 +75,78 @@ func ExecuteJob(ctx context.Context, cfg *Config) error {
...
-			return result, nil
-		}); err != nil {
-			return fmt.Errorf("failed to submit job to worker pool: %w", err)
-		}
-	}
-	// When all of the workers are complete, extract the result values
-	results, err := commitPool.Done(ctx)
+	commitReviewStatuses, err := pooledTransform(ctx, commits,
+		func(commit *Commit) (*CommitReviewStatus, error) {
+			return processCommit(ctx, *commit, gitHubClient), nil
```
verbanicm, line 80: Is there a reason to have the processCommit func deference the commit var? Can we pass the pointer instead?

lock14: It's  a holdover from when those functions were written to use apache beam. Changed to use a pointer instead.

pkg/review/job.go
lock14: It's  a holdover from when those functions were written to use apache beam. Changed to use a pointer instead.

verbanicm (approved):

pkg/review/commit_review_status.go
```diff
@@ -143,12 +143,12 @@ type PageInfo struct {
...
 // PR for the commit targeting the repository's main branch with reviewDecision
 // of 'APPROVED'.
-func processCommit(ctx context.Context, commit Commit, gitHubClient *githubv4.Client) *CommitReviewStatus {
+func processCommit(ctx context.Context, gitHubClient *githubv4.Client, commit *Commit) *CommitReviewStatus {
 	logger := logging.FromContext(ctx)
 	logger.InfoContext(ctx, "process commit", "commit", commit)
 
 	commitReviewStatus := CommitReviewStatus{
-		Commit:         commit,
+		Commit:         *commit,
```
verbanicm, line 151: Can this be a pointer too? Seems like this was probably the reason before also. To avoid the deferencing.

lock14: So the `Commit` struct is embedded in the `CommitReviewStatus` struct. This is because the columns in the `commit_review_status` BigQuery table are a superset of the columns in the commit view.
https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/review/commit_review_status.go#L53-L77

If we change it a pointer I'm not sure how that might afftect how the structs get mapped to BigQuery tables. We can always try it and see what happens.

lock14: I've updated it to a pointer, if this causes problems then I'll revert it later.

pkg/review/job.go
```diff
@@ -81,61 +75,78 @@ func ExecuteJob(ctx context.Context, cfg *Config) error {
...
 
-	// Step 3: Look up break glass issue if necessary.
-	statusPool := workerpool.New[*CommitReviewStatus](&workerpool.Config{
+	return nil
+}
+
+// pooledTransform transforms each input element of type T into an element of type V using the given transform function.
+// The transform is fanned out using a worker pool so that each input element may be processed asynchronously from the
+// others.
+func pooledTransform[T, V any](ctx context.Context, elements []T, transform func(T) (V, error)) ([]V, error) {
```
sethvargo, line 115: I don't understand the desire to abstract this, and the abstraction is not idiomatic. Go does not typically do functional programming patterns, and it's usually a substantial performance overhead over inlining.

pkg/review/job.go
```diff
@@ -81,61 +75,78 @@ func ExecuteJob(ctx context.Context, cfg *Config) error {
...
+	taggedReviewStatuses, err := pooledTransform(ctx, commitReviewStatuses,
+		func(status *CommitReviewStatus) (*CommitReviewStatus, error) {
+			return processReviewStatus(ctx, fetcher, cfg, status), nil
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to process commit review statuses: %w", err)
+	}
+	// statuses that should not be further processed will be nil
+	taggedReviewStatuses = removeNil(taggedReviewStatuses)
```
sethvargo, line 102: Why would any of these return nil? If they do, it'd be better to skip them on the next iteration instead of making an entire copy of the slice with nils removed.

sethvargo (changes requested):

pkg/review/job.go
lock14: There was a bug where the wrong slice was being referenced due to the worker pool code logic being copy/pasted. 

The original code with the bug is here:
https://github.com/abcxyz/github-metrics-aggregator/blob/6caa020e41f833bf7e3cf72d785f7f6742a03f8c/pkg/review/job.go#L132

After the looking at the worker pool code it was clear that there was a lot of boilerplate going on in setting up the pool and accumulating the results. So i figured I would extract the worker pool code into a function in order to avoid this type of error from occurring again. 

I'm not sure how much a performance hit there really is in this case. When the job is invoked, the `pooledTransform` function will be called exactly twice. If this were say a function that would be called many times I could see the argument for inlining being better for performance (and would likely agree that it was the better choice).

pkg/review/job.go
lock14: This is due to the underly processing functions [processCommit](https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/review/commit_review_status.go#L146) and [processReviewStatus](https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/review/commit_review_status.go#L240) both using `nil` as signal that the item should not be further processed.

I agree that copying the entire slice is not the best approach. I will either make a new PR to address this or include a fix for this in the resubmitted PR if we decide to revert this one.

PR #262, opened 2024-05-02
Retag existing image instead of building image without additional layers

mering:
which is more efficient

PR #318, opened 2024-08-13
Update to use new release process

### LINKED ISSUE #315: Update to use new release process
### TL;DR

Now that we've gotten rid of the Beam stuff, we can use the new release process that JVS and other tools have switch to. It's faster and more flexible.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

lock14:
closes #315


pdewilde (approved):
LGTM. May want to add a RELEASING.md or add something in the README about release process.

sethvargo (approved):
LGTM, but we should do a few releases to test it out. Usually we also need to coordinate some changes with our internal repos to pick this up (e.g. WIF permissions). 

Does this clo-se #262 and #263?

.github/workflows/release.yml
```diff
@@ -15,33 +15,158 @@
 name: 'release'
 
 on:
-  push:
-    tags:
-      - 'v*'
+  workflow_run:
```
verbanicm, line 18: Why workflow run and not embedding in this workflow via workflow call. This diverges from existing prior art in jvs. I'm not opposed just want to understand the change.

lock14: The `ci` workflow runs on every push to `main`, if we also run the `release` workflow on pushes to `main` then when we do a release the two workflows will run concurrently. 

if we embed the `ci` worflow in the `release` workflow then both `ci` and `release` will be doing the "same work". There also may or may not be a race condition with the two workflows attempting to push a tag for the same commit sha. 

Since all we care about for the release process is that the image gets built with the head commit sha, I figured it made sense to just wait for the `ci` workflow to complete since it is already doing that work. Then the `release` workflow can just re-use the image the `ci` workflow built.

verbanicm: Do we want to release on every push to main though? I think we only want to release when a tag is generated, not on every push to main. When using the create-release process, a tag is automatically generated by github when the release is created, which should then kick off the release workflow instead.

lock14: If you look at https://github.com/abcxyz/jvs/blob/main/.github/workflows/release.yml it runs on every push to main and checks that the head commit message matches a certain prefix, if not it aborts the workflow. Also, in the new process the `release` workflow is what creates the release tag so we can't trigger it on the release tag being created.

verbanicm: Missed that, thanks!

verbanicm (approved):

lock14:
> LGTM, but we should do a few releases to test it out. Usually we also need to coordinate some changes with our internal repos to pick this up (e.g. WIF permissions).
> 
> Does this clo-se #262 and #263?

I'm not sure, it looks to me like those  are addressing a slightly different issue. I can investigate further to confirm what is needed to close them out.

.github/workflows/release.yml
sethvargo: I'm confused why we're doing something different than what the JVS does?

lock14: if we embed the `ci` workflow inside the `release` workflow and trigger the `release` workflow on pushes to main, then both workflows will run when we do a release and both workflows will build the image and tag it  with the head sha. I was worried this might result in some kind of race condition with pushing the images.

It seemed simpler to just have the `release` workflow trigger after the `ci` workflow as that way the image will already be built and we can just use it inside the `release` workflow`.

PR #324, opened 2024-08-15
Fix bugs in release workflow

lock14:
* Need to read `head_sha` from `workflow_run` object.
* Removed reference to `github.head_ref` as this property is only set if the workflow was triggered from a `pull_request` or `pull_request_target`
* Revert version number due to failed release

Reference Docs:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#github-context


mattrandallbecker (approved):

.github/workflows/release.yml
```diff
@@ -27,11 +27,11 @@ env:
   SOURCE_DOCKER_REPO: 'us-docker.pkg.dev/github-metrics-aggreg-i-64d426/ci-images'
   TARGET_DOCKER_REPO: 'us-docker.pkg.dev/abcxyz-artifacts/docker-images'
   IMAGE_NAME: 'github-metrics-aggregator'
-  SOURCE_TAG: '${{ github.event.head_sha }}'
+  SOURCE_TAG: '${{ github.event.workflow_run.head_sha }}'
```
sethvargo, line 30: Why do we want workflow_run instead of push?

lock14: Yes, hopefully my comment in #318 explains why I went with triggering the release of the cir workflow instead of a push to main. If not, I'm happy to sync and discuss further.

PR #331, opened 2024-09-11
Artifact job - Link to uploaded artifacts in PR comments

Celdir:
Example comment: https://github.com/abcxyz/guardian/pull/406#issuecomment-2344776806

### DIFF
--- a/pkg/artifact/ingest_logs.go
+++ b/pkg/artifact/ingest_logs.go
@@ -28,9 +28,10 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/abcxyz/github-metrics-aggregator/pkg/githubclient"
 	"github.com/abcxyz/pkg/githubauth"
 	"github.com/abcxyz/pkg/logging"
+	"github.com/google/go-github/v61/github"
+	"golang.org/x/oauth2"
 )
 
 // EventRecord maps the columns from the driving BigQuery query
@@ -69,11 +70,9 @@ var errLogsExpired = errors.New("GitHub logs expired")
 
 // logIngester is an object that provides the main processing of the event.
 type logIngester struct {
-	client             *http.Client
-	githubInstallation *githubauth.AppInstallation
-	ghClient           *githubclient.GitHub
-	storage            ObjectWriter
-	bucketName         string
+	ghClient   *github.Client
+	storage    ObjectWriter
+	bucketName string
 }
 
 // NewLogIngester creates a logIngester and initializes the object store, GitHub app and http client.
@@ -94,22 +93,17 @@ func NewLogIngester(ctx context.Context, logsBucketName, gitHubAppID, gitHubInst
 		return nil, fmt.Errorf("failed to get github app installation: %w", err)
 	}
 
-	ghClient, err := githubclient.NewWithPermissions(ctx, gitHubAppID, gitHubInstallID, gitHubPrivateKey, map[string]string{
+	ts := installation.AllReposOAuth2TokenSource(ctx, map[string]string{
 		"actions":       "read",
 		"pull_requests": "write",
 	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to create github client: %w", err)
-	}
+
+	ghClient := github.NewClient(oauth2.NewClient(ctx, ts))
 
 	return &logIngester{
-		storage: store,
-		client: &http.Client{
-			Timeout: 5 * time.Minute,
-		},
-		githubInstallation: installation,
-		ghClient:           ghClient,
-		bucketName:         logsBucketName,
+		storage:    store,
+		ghClient:   ghClient,
+		bucketName: logsBucketName,
 	}, nil
 }
 
@@ -175,10 +169,14 @@ func (f *logIngester) ProcessElement(ctx context.Context, event EventRecord) Art
 // handleMessage is the main event processor. It generates a GitHub token, reads the workflow
 // log files if they exist and persists them to Cloud Storage. It returns the URL to the uploaded storage object.
 func (f *logIngester) handleMessage(ctx context.Context, ghLogsURL, gcsPath string) (string, error) {
-	res, err := f.ghClient.DoRequest(ctx, http.MethodGet, ghLogsURL, nil)
+	req, err := f.ghClient.NewRequest(http.MethodGet, ghLogsURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("error creating GitHub request GET %s: %w", ghLogsURL, err)
+	}
+	res, err := f.ghClient.BareDo(ctx, req)
 	if err != nil {
 		if res == nil {
-			return "", fmt.Errorf("error executing GitHub request: %w", err)
+			return "", fmt.Errorf("error executing GitHub request GET %s: %w", ghLogsURL, err)
 		}
 		// Check for not found conditions. This signals that the logs have expired
 		// and there is nothing that can be done about it.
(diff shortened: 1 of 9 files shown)


pkg/artifact/ingest_logs_test.go
```diff
@@ -231,3 +418,8 @@ func (w *testObjectWriter) Write(ctx context.Context, reader io.Reader, descript
 	w.gotArtifact = string(content)
 	return nil
 }
+
+// toPtr is a helper function to convert a type to a pointer of that same type.
+func toPtr[T any](i T) *T {
```
verbanicm, line 423: nit: can you use our pointer package instead?

https://github.com/abcxyz/pkg/blob/main/pointer/pointer.go#L19

pkg/artifact/ingest_logs.go
```diff
@@ -88,12 +95,21 @@ func NewLogIngester(ctx context.Context, logsBucketName, gitHubAppID, gitHubInst
 		return nil, fmt.Errorf("failed to get github app installation: %w", err)
 	}
 
+	ghClient, err := githubclient.NewWithPermissions(ctx, gitHubAppID, gitHubInstallID, gitHubPrivateKey, map[string]string{
```
verbanicm, line 98: Were mixing the existing impl that uses the token from an installation with generating a github client.

I think we should create a new github client in this func and use it in the existing `handleMessage` func. Instead of creating a raw http request we should use `client.NewRequest` to use the token from the github client.

The permissions should be pull_request:write and actions:read for all repos.

@bradegler  thoughts?

pkg/githubclient/githubclient.go
```diff
@@ -43,6 +43,38 @@ func New(ctx context.Context, appID, rsaPrivateKeyPEM string) (*GitHub, error) {
 	}, nil
 }
 
+func NewWithPermissions(ctx context.Context, appID, gitHubInstallID, rsaPrivateKeyPEM string, permissions map[string]string, authOpts ...githubauth.Option) (*GitHub, error) {
```
verbanicm, line 46: Given my above comment, we wouldnt need this function and the client can be created in the NewLogIngestor func.

pkg/artifact/ingest_logs.go
bradegler: Agreed. We should not mix and match.

pkg/artifact/ingest_logs.go
```diff
@@ -197,3 +220,37 @@ func (f *logIngester) handleMessage(ctx context.Context, repoName, ghLogsURL, gc
...
+		)
+		return nil
+	}
+
+	for _, prNumberStr := range event.PullRequestNumbers {
+		pantheonLogsURL := fmt.Sprintf("https://pantheon.corp.google.com/storage/browser/%s/%s/%s?project=%s", f.bucketName, event.RepositorySlug, event.DeliveryID, f.projectID)
+		comment := fmt.Sprintf("Logs for workflow run [%s](%s) attempt %s uploaded to GCS [here](%s)", event.WorkflowRunID, event.WorkflowURL, event.WorkflowRunAttempt, pantheonLogsURL)
+		prNumber, err := strconv.Atoi(prNumberStr)
+		if err != nil {
+			return fmt.Errorf("error parsing pr number from event payload")
```
bradegler, line 241:
```suggestion
			return fmt.Errorf("error parsing pr number from event payload: %w", err)
```

You are losing the actual error here.

pkg/artifact/ingest_logs.go
```diff
@@ -197,3 +220,37 @@ func (f *logIngester) handleMessage(ctx context.Context, repoName, ghLogsURL, gc
...
+		}
+		resp, err := f.ghClient.CommentPR(ctx, event.OrganizationName, event.RepositoryName, prNumber, comment)
+		if err != nil {
+			return fmt.Errorf("error commenting artifact on pull request: %w", err)
+		}
+		if resp.StatusCode != http.StatusCreated {
+			content, err := io.ReadAll(io.LimitReader(resp.Body, 256_000))
+			if err != nil {
+				return fmt.Errorf("unexpected response status %s for commenting artifact on pull request - failed to read response body", resp.Status)
```
bradegler, line 250: Same as above

```suggestion
				return fmt.Errorf("unexpected response status %s for commenting artifact on pull request - failed to read response body: %w", resp.Status, err)
```

pkg/githubclient/githubclient.go
Celdir: It's worth noting that `githubclient` is also used by the retry server ([example](https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/retry/retry.go#L172)). Previously I had both the retry server and the artifact job using `func New` to create a client that was initialized with the right set of permissions for both, but iirc I was getting some error about the github app not having access to the permissions needed for `ListHookDeliveries` and `RedeliverHookDelivery`.

pkg/artifact/ingest_logs.go
```diff
@@ -197,3 +220,37 @@ func (f *logIngester) handleMessage(ctx context.Context, repoName, ghLogsURL, gc
...
+		logger.InfoContext(
+			ctx,
+			"skipping PR comment for non-successful log ingestion artifact",
+			"delivery_id", event.DeliveryID,
+		)
+		return nil
+	}
+
+	for _, prNumberStr := range event.PullRequestNumbers {
+		pantheonLogsURL := fmt.Sprintf("https://pantheon.corp.google.com/storage/browser/%s/%s/%s?project=%s", f.bucketName, event.RepositorySlug, event.DeliveryID, f.projectID)
```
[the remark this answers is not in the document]
bradegler: You could simplify this by having the code that writes the artifact https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/artifact/storage.go#L49-L73 return the url to the object which is accessible on the `ObjectHandle` https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L1351

Then you don't have to specifically craft this url and if the scheme changes for some reason we won't have an issue. Additionally it removes the need for the project id.

bradegler (changes requested):

pkg/artifact/ingest_logs_test.go
Celdir: done

pkg/artifact/ingest_logs.go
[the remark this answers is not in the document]
Celdir: done

pkg/artifact/ingest_logs.go
Celdir: done

pkg/artifact/ingest_logs.go
Celdir: done

pkg/artifact/ingest_logs.go
Celdir: Refactored `handleMessage` to use the client, PTAL.

pkg/artifact/ingest_logs.go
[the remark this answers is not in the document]
Celdir: So I made this change, but it seems that the `MediaURL` isn't signed and gives an auth error: https://screenshot.googleplex.com/AsGyqmTjKSGss7Q

Should I generate a SignedURL instead?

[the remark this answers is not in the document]
bradegler: Hmm, yes, I guess so. Is there a way to request that from the GCS client?

pkg/githubclient/githubclient.go
verbanicm: I think I was aiming for something like this except generating a github rest client: https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/review/job.go#L40-L60

Celdir: done

pkg/artifact/ingest_logs.go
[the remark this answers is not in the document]
Celdir: Yes but the links would expire in at most 7 days. I went back to using string formatting to generate the cloud console link.

bradegler (approved):

pkg/artifact/ingest_logs.go
```diff
@@ -121,7 +133,9 @@ func (f *logIngester) ProcessElement(ctx context.Context, event EventRecord) Art
 		"delivery_id", event.DeliveryID,
 		"event", event,
 		"result", result)
-	if err := f.handleMessage(ctx, event.RepositoryName, event.LogsURL, gcsPath); err != nil {
+
+	err := f.handleMessage(ctx, event.LogsURL, gcsPath)
+	if err != nil {
```
verbanicm, line 138: nit: prefer single line unless extremely long

```suggestion
	if err := f.handleMessage(ctx, event.LogsURL, gcsPath); err != nil {
```

verbanicm (approved):

pkg/artifact/ingest_logs.go
Celdir: Done

PR #353, opened 2024-12-30
fix: consider the case where deliveries can be an empty list

shankiyani:


pkg/retry/retry.go
```diff
@@ -125,6 +125,12 @@ func (s *Server) handleRetry() http.Handler {
 				return
 			}
 
+			if len(deliveries) == 0 {
```
shankiyani, line 128: Its possible to reach line 137 and experience a NPE. This change should prevent this case.

capri-xiyue (approved):

PR #355, opened 2025-01-12
fix: dont override checkpoint if list deliveries is empty

shankiyani:


pkg/retry/retry.go
```diff
@@ -225,8 +226,11 @@ func (s *Server) handleRetry() http.Handler {
 		}
 
 		// advance the checkpoint to the first entry read on this run to avoid
-		// redundant processing
-		newCheckpoint = firstCheckpoint
+		// redundant processing, handle edge case where the list deliveries API
+		// does not return any events
+		if firstCheckpoint != "" {
+			newCheckpoint = firstCheckpoint
```
sethvargo, line 232: We should add a log message (probably debug, but maybe warn)

PR #360, opened 2025-01-23
Add checks to avoid referencing "null" values

ian-shafer:


terraform/outputs.tf
```diff
@@ -117,6 +117,7 @@ output "github_metrics_looker_studio_report_link" {
 }
 
 output "artifacts_job" {
+  count = var.leech.enabled ? 1 : 0
```
gjonathanhong, line 120: `count` keyword isn't accepted here. Maybe we could use the `try` keyword for now?

e.g.
```
value = {
    job_id                 = try(module.leech[0].job_id, null)
    job_name               = try(module.leech[0].job_name, null)
    service_account_name   = try(module.leech[0].google_service_account.name, null)
    service_account_email  = try(module.leech[0].google_service_account.email, null)
    service_account_member = try(module.leech[0].google_service_account.member, null)
}
```

ian-shafer: Good catch. I also looked at using `precondition` but that seems to just emit errors. Done.

gjonathanhong (approved):

PR #372, opened 2025-04-21
fix: lint problems

dcreey:
Temporarily adds the lint.yml from abcxyz/actions, then fixes lint problems, then removes the lint.yml. This repo will be added to the org ruleset so it gets linted after this merges.


sethvargo (commented):
We may want to disable that actionlint rule.

dcreey:
> We may want to disable that actionlint rule.

Agree. I'll resolve it globally

verbanicm (approved):

PR #377, opened 2025-06-10
Add enterprise view

ruizhg:
Adds a enterprises view with a unique list of enterprise IDs, sourced from the raw_events table.

This uses the existing module automation by adding:
- enterprise_events.sql
- enterprises.sql

### DIFF
--- a/terraform/modules/bigquery_metrics_views/data/bq_views/events/enterprise_events.sql
+++ b/terraform/modules/bigquery_metrics_views/data/bq_views/events/enterprise_events.sql
@@ -1,0 +1,25 @@
+
+-- Copyright 2025 The Authors (see AUTHORS file)
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


verbanicm:
@pdewilde can you weigh in here?

terraform/modules/bigquery_metrics_views/data/bq_views/events/enterprise_events.sql
pdewilde, line 1: A few thoughts:

1.) This should be implemented as a TVF rather than a view.

https://cloud.google.com/bigquery/docs/table-functions

The reason is a bit silly, but clustering and partitioning are not able to be used (at least not exposed in billing usage) over view bounaries.

This is a bit of a toy example, as currently date isn't being exposed here. Imagine I am interested in events in the last 10 hours. I could query:

```sql
-- Note, I could be misremembering and this requires another layer of nested views to be pathalogical, I would need to test it
SELECT *
FROM enterprise_events e
WHERE e.date > current_date
```


But BigQuery would bill you for scanning all data in the table even though you are only interested in a tiny bit of data. This is because BigQuery first materializes the entire view, then applies the filter. A good DB optimizer wouldn't do that, but this is BigQuery...

If you use a TVF which accepts date as an argument, you can then choose to only scan the part you are interested. You can have nested TVFs which each pass down filters on clustered and partitioned columns.

2.) I would recommend exposing all fields from raw_events table. While your enterprises view doesn't use them, others may want to the future, and there isn't really an extra cost by including them.

3.) An alternative would be adding this to the unique events TVF we already have https://github.com/abcxyz/github-metrics-aggregator/blob/main/terraform/bigquery.tf#L453 or using that as the base and just parsing another field out of the payload. This would allow for more usable data if someone wants to use it in the future.

pdewilde (approved):

PR #378, opened 2025-06-18
Enable github_metrics_dashboard

ruizhg:
To view the Looker Studio dashboard, enable it and add the viewers.

I have two questions:
1. I was exploring how to set the "github_metrics_dashboard" variable, but I haven't found a good way to do so. 
2. For viewers, I simply used my own email for testing.

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -425,8 +425,8 @@ variable "github_metrics_dashboard" {
     viewers          = optional(list(string), [])
   })
   default = {
-    enabled = false
-    viewers = []
+    enabled = true
+    viewers = [user:ruizhg@gmail.com]
   }
 }
 


gjonathanhong:
> I have two questions:
> 
> 1. I was exploring how to set the "github_metrics_dashboard" variable, but I haven't found a good way to do so.

Set these values in the repository that this module is actuated/applied, not here. What you're doing here is modifying the Terraform module, so anyone depending on this module may pick up this change if they are pinned to main.

gjonathanhong:
No, the `github_metrics_dashboard` is a variable of the module. You just need to set the value of the variable in the top-level module (in the repository that it's applied in).

gjonathanhong:
> No, the `github_metrics_dashboard` is a variable of the module. You just need to set the value of the variable in the top-level module (in the repository that it's applied in).

It will look something like:
```
module "gma" {
   source = "git::https://github.com/abcxyz/github-metrics-aggregator.git//terraform?ref=<123456789>"

   github_metrics_dashboard = {
      enabled = true
      viewers = [ ... ]
   }
}
```

ruizhg:
https://github.com/abcxyz/infra-gcp/pull/2173 is the correct way to enable the looker studio dashboard.

PR #380, opened 2025-07-17
feat: support GHES instance

sqin2019:
Add support for GHES github app and client.

TODOs:
- terraform module update
- review job for ghes
- kms key signer with github private key

### DIFF
--- a/pkg/review/config.go
+++ b/pkg/review/config.go
@@ -17,6 +17,7 @@ package review
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/sethvargo/go-envconfig"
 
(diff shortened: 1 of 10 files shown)


pkg/review/config.go
```diff
@@ -27,9 +28,10 @@ import (
 // Config defines the set of environment variables required
 // for running the artifact job.
 type Config struct {
-	GitHubAppID            string `env:"GITHUB_APP_ID,required"`             // The GitHub App ID
-	GitHubInstallID        string `env:"GITHUB_INSTALL_ID,required"`         // The provisioned GitHub App Installation reference
-	GitHubPrivateKeySecret string `env:"GITHUB_PRIVATE_KEY_SECRET,required"` // The secret name & version containing the GitHub App private key
+	GitHubEnterpriseServerURL string `env:"GITHUB_ENTERPRISE_SERVER_URL"`       // The GitHub Enterprise Server instance URL, format "https://[hostname]"
+	GitHubAppID               string `env:"GITHUB_APP_ID,required"`             // The GitHub App ID
+	GitHubInstallID           string `env:"GITHUB_INSTALL_ID,required"`         // The provisioned GitHub App Installation reference
```
bradegler, line 33: This is going to need to be changed to look up the install id instead of having it passed in from the command line / environment. 

You'll have an installation ID per org and we can't have an instance of GMA per org. You'll need to get the installation id for each org/repo somehow like what we do in minty: https://github.com/abcxyz/github-token-minter/blob/main/pkg/server/source/github.go#L71

bradegler (changes requested):

pkg/review/config.go
sqin2019: Make sense! I will do it in a separate CL since it requires some refactor.

drevell (approved):
LGTM for go readability, I defer to Brad for GitHub app logic

bradegler (approved):

pkg/review/config.go
```diff
@@ -30,7 +30,6 @@ import (
 type Config struct {
 	GitHubEnterpriseServerURL string `env:"GITHUB_ENTERPRISE_SERVER_URL"`       // The GitHub Enterprise Server instance URL, format "https://[hostname]"
 	GitHubAppID               string `env:"GITHUB_APP_ID,required"`             // The GitHub App ID
-	GitHubInstallID           string `env:"GITHUB_INSTALL_ID,required"`         // The provisioned GitHub App Installation reference
```
sqin2019: @bradegler ended up adding the change as it turned out not much refactor needed, please take another look, though it does not include updating the terraform modules etc.

bradegler (approved):

pkg/review/config.go
bradegler: This looks good. Are you planning on running the artifacts job too at some point? That will need a similar change.

sqin2019: Not for GHSS, but I can take a look at it now.

PR #396, opened 2025-08-04
feat: support github app private key stored in kms

sqin2019:
support optional kms integration for github app private key


pkg/artifact/config.go
```diff
@@ -53,8 +54,8 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("GITHUB_APP_ID is required")
 	}
 
-	if cfg.GitHubPrivateKeySecret == "" {
-		return fmt.Errorf("GITHUB_PRIVATE_KEY_SECRET is required")
+	if cfg.GitHubPrivateKeySecret == "" && cfg.GitHubPrivateKeyKMSKeyID == "" {
```
sethvargo, line 57: Should these be mutually exclusive (an error to provide both)?

sethvargo (changes requested):
Hi @sqin2019 - this is a lot of duplication, and I'm not sure the coupling makes sense. The GitHub pieces really shouldn't care about the kind of signer.

I thought we updated the interfaces to accept a `crypto.Signer` (https://github.com/abcxyz/pkg/blob/main/githubauth/app.go#L94), which greatly reduces the complexity here.

/cc @bradegler

pkg/githubclient/githubclient.go
```diff
@@ -28,17 +31,11 @@ type GitHub struct {
...
-// NewGitHubEnterpriseClient creates a new instance of a GitHub client (for
-// enterprise if enterpriseURL is non-empty).
-func NewGitHubEnterpriseClient(ctx context.Context, enterpriseURL, appID, rsaPrivateKeyPEM string) (*GitHub, error) {
-	app, err := NewGitHubApp(ctx, enterpriseURL, appID, rsaPrivateKeyPEM)
+// New creates a new GitHub client.
+func New(ctx context.Context, signer crypto.Signer, enterpriseURL, appID string) (*GitHub, error) {
+	app, err := NewGitHubApp(ctx, signer, enterpriseURL, appID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create github app: %w", err)
+		return nil, fmt.Errorf("failed to create github app from kms: %w", err)
```
sethvargo, line 38: I think this is stale now?

pkg/githubclient/githubclient.go
```diff
@@ -49,12 +46,7 @@ func NewGitHubEnterpriseClient(ctx context.Context, enterpriseURL, appID, rsaPri
 	return &GitHub{client: client}, nil
 }
 
-func NewGitHubApp(ctx context.Context, enterpriseURL, appID, rsaPrivateKeyPEM string) (*githubauth.App, error) {
-	signer, err := githubauth.NewPrivateKeySigner(rsaPrivateKeyPEM)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create private key signer: %w", err)
-	}
-
+func NewGitHubApp(ctx context.Context, signer crypto.Signer, enterpriseURL, appID string) (*githubauth.App, error) {
```
sethvargo, line 49: How is this different from `New` now?

sethvargo (commented):
This looks better, but I'm still confused why we have so much initialization duplication across all these packages?

sqin2019:
> This looks better, but I'm still confused why we have so much initialization duplication across all these packages?

GMA services/jobs use different GitHub authentication and API, retry service uses github app (rest api), review(rest api) and artifact job (graphql) uses github app installation.

sethvargo:
> > This looks better, but I'm still confused why we have so much initialization duplication across all these packages?
> 
> GMA services/jobs use different GitHub authentication and API, retry service uses github app (rest api), review(rest api) and artifact job (graphql) uses github app installation.

It's still shareable. See https://github.com/abcxyz/github-metrics-aggregator/pull/397 which is a PR against this.

PR #408, opened 2025-09-22
Retry serivce cloud run job exectuion limit 5 minutes -> 29 minutes 45 seconds

pdewilde:
The scheduler duration had been set to 30 minutes, but the cloud run job definition was set to timeout after 5 minutes (default).

This consumes a new version of terraform-modules that exposes the timeout_seconds field and sets it to just below that maximum timeout of the scheduler.

### DIFF
--- a/terraform/service_retry.tf
+++ b/terraform/service_retry.tf
@@ -137,7 +137,7 @@ module "retry_cloud_run" {
     },
   }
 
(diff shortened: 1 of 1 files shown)


pdewilde:
Ideally we will move away from using a http request to trigger retry jobs

bradegler (approved):

PR #417, opened 2025-09-25
feat: refactored retry service into a cloud run job

### LINKED ISSUE #413: Change Retry Service from an http service to a job
Retry Service is currently triggered via a blocking HTTP call from Cloud Scheduler.

This is problematic as cloud scheduler has a max limit of 30 minutes. Also, blocking http calls ususally aren't used for long-running jobs.

We should switch to a cloud run job based workflow. It can be triggered like so: https://cloud.google.com/run/docs/execute/jobs-on-schedule#terraform

bradegler:
closes #413

This PR refactors the retry service from a Cloud Run service to a Cloud Run job. This is a more appropriate and efficient model for the retry mechanism, which runs periodically to find and retry failed webhook deliveries.

Key changes include:

- Application Logic: The HTTP server code in pkg/retry has been removed and replaced with a ExecuteJob function that encapsulates the retry logic. The command-line interface has been updated to execute this job.
- Terraform: The infrastructure in terraform/service_retry.tf has been updated to deploy a google_cloud_run_v2_job instead of a google_cloud_run_service. This includes the necessary IAM roles and a google_cloud_scheduler_job to trigger the job.
- GitHub Actions: The CI/CD workflow in .github/workflows/ci.yml has been updated to remove the deployment of the old retry service. This workflow will be reworked to deploy the CR job once the terraform can be applied in those environments.
- Testing: Unit tests have been updated to reflect the new job-based architecture.

This change simplifies the architecture and reduces the resource footprint of the retry mechanism.

want_lgtm=all

### DIFF
--- a/terraform/modules/retry/variables.tf
+++ b/terraform/modules/retry/variables.tf
@@ -126,7 +126,7 @@ variable "github_enterprise_server_url" {
 }
 
 variable "timeout" {
-  description = "The task timeout setting see: https://cloud.google.com/run/docs/configuring/task-timeout#set_task_timeout. Defaults to 1 hour"
+  description = "The task timeout setting see: https://cloud.google.com/run/docs/configuring/task-timeout#set_task_timeout. Defaults to 10 minutes"
   type        = string
-  default     = "3600s"
+  default     = "600s"
 }
--- a/terraform/service_retry.tf
+++ b/terraform/service_retry.tf
@@ -28,6 +28,7 @@ module "retry_job" {
   github_private_key_secret_id      = "github-private-key"
   github_private_key_secret_version = "latest"
   scheduler_cron                    = var.retry_job_schedue
+  timeout                           = var.retry_job_timeout
   retry_job_iam = {
     admins     = toset(var.retry_service_iam.admins)
     developers = toset(concat(var.retry_service_iam.developers, [var.automation_service_account_member]))
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -577,6 +577,12 @@ variable "webhook_max_instances" {
 
 variable "retry_job_schedue" {
   type        = string
-  default     = "*/30 * * * *"
-  description = "Frequencey to run the retry job. Follows standard cron syntax. Defaults to every 30 minutes."
+  default     = "0 * * * *"
+  description = "Frequencey to run the retry job. Follows standard cron syntax. Defaults to every hour."
+}
+
+variable "retry_job_timeout" {
+  description = "The task timeout setting see: https://cloud.google.com/run/docs/configuring/task-timeout#set_task_timeout. Defaults to 45 minutes"
+  type        = string
+  default     = "2700s"
 }


terraform/modules/retry/main.tf
```diff
@@ -0,0 +1,209 @@
...
+  project = var.project_id
+
+  name     = var.job_name
+  location = var.region
+
+  template {
+    parallelism = 0
+    task_count  = 1
+
+    template {
```
pdewilde, line 38: https://cloud.google.com/run/docs/configuring/task-timeout

You will need to define timeout. The default is 5 minutes which will be far too short for processing large volumes.

terraform/modules/retry/variables.tf
```diff
@@ -0,0 +1,126 @@
...
+}
+
+variable "checkpoint_table_id" {
+  description = "The BigQuery checkpoint table id to create."
+  type        = string
+  nullable    = false
+}
+
+variable "scheduler_deadline_duration" {
+  description = "The deadline for job attempts in seconds. If the request handler does not respond by this deadline then the request is cancelled and the attempt is marked as a DEADLINE_EXCEEDED failure. Defaults to 30 minutes."
```
pdewilde, line 88:
```suggestion
  description = "The deadline for job attempts in seconds. If the request handler does not respond by this deadline then the request is cancelled and the attempt is marked as a DEADLINE_EXCEEDED failure. Defaults to 3 minutes."
```

terraform/modules/retry/variables.tf
```diff
@@ -0,0 +1,126 @@
...
+}
+
+variable "scheduler_retry_limit" {
+  description = "Number of times Cloud Scheduler will retry the job when it "
+  type        = string
+  default     = "0"
+}
+
+variable "image" {
+  description = "Cloud Run service image for github-metrics-aggregator and server entrypoints."
```
pdewilde, line 112: > github-metrics-aggregator and server entrypoints. 

Shouldn't that be the retry endpoint?

terraform/service_retry.tf
```diff
@@ -12,15 +12,29 @@
...
+  region                            = var.region
+  image                             = var.image
+  dataset_id                        = var.dataset_id
+  events_table_id                   = var.events_table_id
+  checkpoint_table_id               = var.checkpoint_table_id
+  bucket_name                       = google_storage_bucket.retry_lock.name
+  github_app_id                     = var.github_app_id
+  github_private_key_secret_id      = "github-private-key"
+  github_private_key_secret_version = "latest"
+  scheduler_cron                    = "*/30 * * * *"
```
pdewilde, line 30: 1. I would reccomend making this configurable
2. I think we will want to to run every few hours, otherwise we risk the scenario of overlapping jobs if we are recovering from an outage.

terraform/modules/retry/main.tf
bradegler: Ack. Added a var for this.

terraform/variables.tf
```diff
@@ -574,3 +574,9 @@ variable "webhook_max_instances" {
   description = "The max number of instances for the Webhook Cloud Run service (defaults to '10')."
 
 }
+
+variable "retry_job_schedue" {
+  type        = string
+  default     = "*/30 * * * *"
+  description = "Frequencey to run the retry job. Follows standard cron syntax. Defaults to every 30 minutes."
```
pdewilde, line 581: I think the default task timeout should be less than the default retry job schedule. While it shouldn't be problematic to be running multiple jobs at once due to idempotency checks, I do think its more likely we high rate limits.

pkg/retry/job.go
```diff
@@ -0,0 +1,221 @@
...
+
+	failedEventCount := len(failedEventsHistory)
+	for i := failedEventCount - 1; failedEventCount > 0 && i >= 0; i-- {
+		eventIdentifier := failedEventsHistory[i]
+		if err := githubClient.RedeliverEvent(ctx, eventIdentifier.eventID); err != nil {
+			var acceptedErr *github.AcceptedError
+			if !errors.As(err, &acceptedErr) {
+				exists, deliveryErr := datastore.DeliveryEventExists(ctx, cfg.EventsTableID, eventIdentifier.guid)
+				if deliveryErr != nil {
+					if newCheckpoint != prevCheckpoint {
```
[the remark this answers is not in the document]
bradegler: I'll get this in the next one. This logic is about to change and I'll be sure to break it up so its easier to manage and test.

PR #439, opened 2025-12-17
Fix: Add error handling for 401 Bad Credentials error from GitHub

jessikad-google:
Refactored ExecuteJob to handle 401 Bad Credentials errors from GitHub.
Added a retry loop that refreshes the GitHub client upon receiving a 401 and retries the request once.
Removed time-based token refresh logic and the Now function from 
RetryClientOptions

Test
Updated TestExecuteJob_TokenRefresh to simulate a 401 error and verify the error-based refresh mechanism.

### DIFF
--- a/pkg/retry/job.go
+++ b/pkg/retry/job.go
@@ -64,7 +64,6 @@ type RetryClientOptions struct {
 }
 
 // ExecuteJob runs the retry job to find and retry failed webhook events.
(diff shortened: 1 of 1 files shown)


jessikad-google:
Sounds good! I was still getting the timing error yesterday too but both are 401 anyway. I'll work on that.

bradegler (approved):

PR #442, opened 2025-12-22
feat: add a service that will forward messages from one topic to another

bradegler:
### Description
This PR implements the new Relay Service, a component designed to ingest PubSub events, enrich them with GitHub metadata (Enterprise, Organization, Repository names/IDs), and forward them to a configured destination topic.

This will be useful to feed a secondary topic in a shared project, similar to the shared BQ instance. We can then allow partners to subscribe to this centralized topic that will contain events for all deployments.

This change includes the core Go implementation, CLI command, comprehensive unit tests, and the necessary Terraform configuration for deployment.

### Key Changes

1. Core Service Implementation (pkg/relay)
pkg/relay/relay.go: Implemented the handleRelay HTTP handler which:
Accepts PubSub push messages.
Unmarshals the underlying event payload.
Enriches the event with Enterprise, Organization, and Repository metadata.
Publishes the enriched event to a destination PubSub topic.
pkg/relay/server.go: Implemented the server setup with a Messenger interface to abstract the PubSub client for better testability.
pkg/relay/config.go: Defined configuration structure and validation logic.

2. CLI Integration (pkg/cli)
pkg/cli/relay.go: Added the relay command to the CLI, wiring up the configuration, server initialization, and signal handling.
cmd/github-metrics-aggregator/main.go (implied): Registered the new command.

3. Terraform Deployment (terraform/)
Service Definition: Added terraform/service_relay.tf to define the Cloud Run service and a dedicated Service Account.
Subscription: Configured a PubSub Subscription (google_pubsub_subscription) with push configuration to route events from the default ingestion topic to the Relay service.
Includes OIDC authentication using a dedicated subscription identity.
Includes Dead Letter Policy.
Variables: Added necessary variables (enable_relay_service, relay_topic_id, relay_project_id) to terraform/variables.tf. The service is disabled by default.

4. Testing & Quality
Unit Tests: Added comprehensive tests in pkg/relay/relay_test.go covering success scenarios, malformed requests, and error handling.
Linting: Fixed various lint errors including errchkjson (checked map marshaling), forcetypeassert (safe type assertions), and protogetter (used generated protobuf getters).
Documentation: Added GoDoc comments for all exported types and functions.

### Verification
Automated Tests
go test ./pkg/relay/... passed with 100% coverage.
terraform validate passed successfully.

### Manual Verification
Verified that the Terraform configuration uses correct project and region attributes for the Cloud Run service IAM bindings.
Verified safe JSON unmarshaling and error handling in the enricher logic.

### DIFF
--- a/pkg/pubsub/pubsub.go
+++ b/pkg/pubsub/pubsub.go
@@ -26,7 +26,6 @@ import (
 // Messenger defines the interface for sending messages to a relay destination.
 type Messenger interface {
 	Send(ctx context.Context, msg []byte, attrs map[string]string) error
-	Close() error
 }
 
 // PubSubMessenger implements the Messenger interface for Google Cloud pubsub.
@@ -65,15 +64,6 @@ func (p *PubSubMessenger) Send(ctx context.Context, msg []byte, attrs map[string
 
 	if _, err := result.Get(ctx); err != nil {
 		return fmt.Errorf("pubsub: failed to get result: %w", err)
-	}
-	return nil
-}
-
-// Close handles the graceful shutdown of the pubsub client.
-func (p *PubSubMessenger) Close() error {
-	p.topic.Stop()
-	if err := p.client.Close(); err != nil {
-		return fmt.Errorf("failed to close pubsub client: %w", err)
 	}
 	return nil
 }
--- a/pkg/relay/enricher.go
+++ b/pkg/relay/enricher.go
@@ -1,0 +1,113 @@
+// Copyright 2026 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package relay
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	protos "github.com/abcxyz/github-metrics-aggregator/protos/pubsub_schemas"
+	"github.com/abcxyz/pkg/logging"
+)
+
+type MessageEnricher interface {
+	Enrich(ctx context.Context, req []byte) ([]byte, map[string]string, error)
+}
+
+func NewDefaultMessageEnricher() MessageEnricher {
+	return &defaultMessageEnricher{}
+}
+
+type defaultMessageEnricher struct{}
+
+func (d *defaultMessageEnricher) Enrich(ctx context.Context, req []byte) ([]byte, map[string]string, error) {
+	logger := logging.FromContext(ctx)
+	var m wrappedMessage
+	if err := json.Unmarshal(req, &m); err != nil {
+		logger.ErrorContext(ctx, "failed to decode request body", "error", err)
+		return nil, nil, fmt.Errorf("failed to decode request body: %w", err)
+	}
+
+	logger.DebugContext(ctx, "received pubsub message",
+		"subscription", m.Subscription,
+		"message_id", m.Message.ID,
+		"data", string(m.Message.Data),
+		"attributes", m.Message.Attributes,
+	)
+
+	var event protos.Event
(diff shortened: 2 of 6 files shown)


pdewilde:
An even simpler option may be to use SMT's https://docs.cloud.google.com/pubsub/docs/smts/udfs-overview

pkg/cli/relay.go
```diff
@@ -0,0 +1,103 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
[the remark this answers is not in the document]
bradegler: 👍

pdewilde:
We discussed this a bit more. Really we are trying to do multiple things:

1. Take multiple pub/sub topics and combine them into a single pub/sub topic. The motivation is we have multiple installations of GMA which write to separate topics, but for a consumer it would be difficult to manage multiple subscriptions, especially if we were to change what topics we were producing on.

    A relay service makes sense here. It is simple to write, cheap to deploy, and fits in with our existing architecture.

2. Extract data from the payload. I think we need more thought about how we want this to look and what usecases we see for this.

GMA's webhook produces a pub/sub message with a json body, but the "payload" json from the webhook is just stored as a string. There are a couple ways we could improve this:

- unmarshal the whole body and make the payload field of the pub/sub message a json object rather than a string (we may want to leave the original string in the case of a marshal error, or we could just send to a DLQ)
- unmarshal the body but only promote select fields (enterprise, org, repo, ect) to the top level json object, leaving the payload as a string
- unmarshal the body and put select fields as [attributes](https://docs.cloud.google.com/pubsub/docs/publisher#using-attributes) allowing [filtering](https://docs.cloud.google.com/pubsub/docs/subscription-message-filter) on the subscription side 

 There are a few options here:

- Use a SMT, which allows us to define a UDF (js) that we can use to unmarshal json and pull fields out. This can be used for both data extraction, as well as filtering. You do double the price of processing a message, but its still very cheap ($80 for TiB vs $40 for Tib).
- Make this unmarshalling/enrichment part of the relay service
- Make this unmarshaling/enrichment part of the webhook service

----------------

I could see an argument for making 1.) a simple service that only does the relay. There would be very little config (just what topic to push to) and it would make a nice reusuable open-source service. This would mean additional ci/cd overhead and project management overhead (unless we left ithere)

Where 2.) lives I think is up for debate. I could see strong arugments for putting it in webhook, relay, or SMT. Webhook would keep the relay simple as well as not have another component (smt udf). Relay would prevent SMT and keep webhook simple, at the cost of more logic. SMT is nice because we don't need to manage anything besides providing a UDF, and customers could add their own UDF to their own subscriptions following the same pattern as us.

pkg/pubsub/pubsub.go
```diff
@@ -32,7 +39,7 @@ type PubSubMessenger struct {
 }
 
 // NewPubSubMessenger creates a new instance of the PubSubMessenger.
-func NewPubSubMessenger(ctx context.Context, projectID, topicID string, opts ...option.ClientOption) (*PubSubMessenger, error) {
+func NewPubSubMessenger(ctx context.Context, projectID, topicID string, opts ...option.ClientOption) (Messenger, error) {
```
pdewilde: We need to change the timeout of the topic [topic.PublishSettings.Timeout])https://pkg.go.dev/cloud.google.com/go/pubsub#PublishSettings)

The default is 60 seconds https://docs.cloud.google.com/pubsub/docs/publish-best-practices#retry-requests but  we probably want to reduce that to something below the least time for a message.

I would recommend setting the lease time to 6 seconds, publish timeout to 4 seconds, and max http timeout for the cloud run service to 10 seconds (or possibly stricter).

The important part is that:
publishTimeout<leaseTime
leaseTimec<loudRunTimeout

terraform/service_relay.tf
```diff
@@ -0,0 +1,106 @@
...
+resource "google_service_account" "relay_run_service_account" {
+  count = var.enable_relay_service ? 1 : 0
+
+  project = data.google_project.default.project_id
+
+  account_id   = "${var.prefix_name}-relay-sa"
+  display_name = "${var.prefix_name}-relay-sa Cloud Run Service Account"
+}
+
+module "relay_cloud_run" {
```
pdewilde: A few thoughts here:

1.  We should probably tune the timeout way down (see my comments in the pubsub package)
2.  We really should modernize and make a cloud_run_v2 module
3.  Our module just uses the default concurrency limit (80 concurrent requests for >=1 cpu). Might be worth benchmarking cpu/ram usage at this concurrency level and adjust as needed (we should expose a variable for this).

shankiyani (approved):

